PR #585, opened 2025-02-06
Add option to capture Class<T> type of Context.Key

mtdowling:
Context.Key now allows you to optionally provide the Class<T> that's contained in the Key. This makes it possible for Context.Key to be adapted to libraries that provide similar typed attributes but require the Class<T> to be available via a getter of the attribute. However, some libraries don't expose Class<T>. To allow smithy-java compatibility with both kinds of libraries, providing a Class<T> when creating a Context.Key is optional.

This allows smithy-java to interop with both kinds of libraries, though it would not be possible to adapt a context key from a library that doesn't expose Class<T> for use with a library that requires Class<T>.

All publicly exposed Context.Keys in smithy-java have been updated to define the Class<T> of the key.

*Issue #, if available:*

*Description of changes:*


mtdowling:
Closing for now

PR #588, opened 2025-02-07
Add our own exceptions for ClientTransport

mtdowling:
To avoid directly exposing the exceptions of the underlying transport of a ClientTransport, which inadvertently allows for coupling of caller logic to a specific transport, we now provide a set of transport exceptions that extend from CallException.

ClienTransports are expected to only throw a subtype of CallException or ClienTransportException, and to wrap other exceptions. ClientPipeline has been updated to also perform this remapping in case a transport either is implemented incorrectly or tries to work around this mapping (which again allows for coupling to a specific transport).

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/client/client-core/src/main/java/software/amazon/smithy/java/client/core/ClientTransport.java
+++ b/client/client-core/src/main/java/software/amazon/smithy/java/client/core/ClientTransport.java
@@ -5,13 +5,20 @@
 
 package software.amazon.smithy.java.client.core;
 
-import java.io.IOException;
-import java.io.InterruptedIOException;
 import java.net.ConnectException;
+import java.net.ProtocolException;
 import java.net.SocketException;
 import java.net.SocketTimeoutException;
 import java.util.concurrent.CompletableFuture;
+import javax.net.ssl.SSLException;
+import software.amazon.smithy.java.client.core.error.ConnectTimeoutException;
+import software.amazon.smithy.java.client.core.error.TlsException;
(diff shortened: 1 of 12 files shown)


adwsingh (approved):

client/client-core/src/main/java/software/amazon/smithy/java/client/core/ClientTransport.java
```diff
@@ -63,4 +58,32 @@ public interface ClientTransport<RequestT, ResponseT> extends ClientPlugin {
...
+    /**
+     * Remaps a thrown exception to an appropriate {@link TransportException} or {@link CallException}.
+     *
+     * <p>This method attempts to map built-in JDK exceptions to the appropriate subclass of
+     * {@code TransportException}.
+     *
+     * @param e Exception to map to a {@link TransportException} or subclass.
+     * @return the remapped exception. A given {@code CallException} or {@code TransportException} is returned as-is.
+     */
+    static CallException remapExceptions(Throwable e) {
```
hpmellema, line 71: Thought: I wonder if a transport should require an exception mapping method as part of it's interface given that Apache HTTP/Netty HTTP/MQTT transports are all going to need to do some sort of mapping.

mtdowling: I think it's fine to leave that off as a requirement.

PR #620, opened 2025-03-27
smithy-call

franco2002lu:
Initial working version of smithy-call, powered by GraalVM. I have listed a few things to address before we merge. Please feel free to leave any suggestions.

1. Document usage in a README.md.
2. Find a better way to include the BASE_RESOURCE_FILES. I currently re-attached them in the same submodule to make the calls work, but I'd like to move away from that.
3. Better protocol-input handling.
4. Output formatting.


cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,253 @@
...
+import software.amazon.smithy.java.json.JsonCodec;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.shapes.ShapeId;
+
+@Command(name = "smithy-call", mixinStandardHelpOptions = true, version = "1.0",
+        description = "Smithy Java CLI")
+public class SmithyCall implements Callable<Integer> {
+    private static final JsonCodec CODEC = JsonCodec.builder().build();
+
+    private static final String AWS_JSON = "awsjson";
```
mtdowling, line 40: This should probably be an enum

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,253 @@
...
+                System.err.println("Resource not found: " + smithyFile);
+            }
+        }
+
+        // Add model files
+        try {
+            assembler.addImport(directoryPath);
+            return assembler.assemble().unwrap();
+        } catch (Exception e) {
+            throw new IOException("Failed to assemble model from directory: " + e, e);
```
mtdowling, line 146: Do you event need to wrap here?

cli/build.gradle.kts
```diff
@@ -0,0 +1,64 @@
+plugins {
+    `java-library`
+    application
+    id("org.graalvm.buildtools.native") version "0.10.3"
+}
+
+dependencies {
```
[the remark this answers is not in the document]
mtdowling: You need a waiters dependency too if this would ever work with AWS models.

You maybe want to add an `--aut` flag to the CLI to ignore unknown traits at some point. That can wait til later, but the CLI will fail right now if a trait isn't defined.

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,253 @@
...
+    private String url;
+
+    @Option(names = { "-p", "--protocol" }, description = "Communication protocol to use (options: awsjson, rpcv2-cbor, restjson, and restxml)")
+    private String protocol;
+
+    @Option(names = { "-a", "--auth" }, description = "Authentication method to use")
+    private String authType;
+
+    @Option(names = { "--aws-region" }, description = "AWS region for SigV4 authentication")
+    private String awsRegion;
```
haydenbaker, line 77: Hmmm... I wonder if there is a better way to support parameters like this. Not everyone is AWS, so it doesn't make sense to me to have this in here. Same for auth (though that's not obviously not tied to AWS, but its options may be [sigv4]).

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,253 @@
...
+            String content = Files.readString(Path.of(inputPath), StandardCharsets.UTF_8);
+            inputDocument = CODEC.createDeserializer(content.getBytes(StandardCharsets.UTF_8)).readDocument();
+        }
+
+        return client.call(operation, inputDocument).asObject();
+    }
+
+    private void logError(String message, Exception e) {
+        System.err.println(message + ": " + e.getMessage());
+    }
```
haydenbaker, line 252: Let's have a debug/verbose CLI option too. Any particular reason we're not using `java.util.logging.Logger`?

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,253 @@
...
+            "aws.protocols.smithy"
+    };
+
+    @Parameters(index = "0", description = "Service Name")
+    private String service;
+
+    @Parameters(index = "1", description = "Name of the operation to perform on the service", arity = "0..1")
+    private String operation;
+
+    @Option(names = { "-m", "--model-path" }, description = "Path to the .smithy service model file", required = true)
```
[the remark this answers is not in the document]
franco2002lu: It currently does support a directory of model files, I should have updated the description here. I wonder what multi-file support would look like - does the user simply provide a list of of files under one option?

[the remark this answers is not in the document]
mtdowling: Ah cool. For multiple files, I would assume it would be repeated arguments, but I'm not sure what picocli does. `-m foo.smithy -m bar.smithy -b baz.smithy`

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
franco2002lu: I agree, I can make the change so that --aws-region can only be specified by the user if the authType is sigv4. This seems like a duct tape style fix tho, I will be thinking of a better approach for user experience.

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
franco2002lu: You are right, this is redundant

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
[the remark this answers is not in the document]
franco2002lu: Got it, I will look into this

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,253 @@
...
+
+        Document inputDocument;
+        if (input != null) {
+            inputDocument = CODEC.createDeserializer(input.getBytes(StandardCharsets.UTF_8)).readDocument();
+        } else {
+            String content = Files.readString(Path.of(inputPath), StandardCharsets.UTF_8);
+            inputDocument = CODEC.createDeserializer(content.getBytes(StandardCharsets.UTF_8)).readDocument();
+        }
+
+        return client.call(operation, inputDocument).asObject();
```
mtdowling, line 247: I'd return the document as-is, then serialize the document using CODEC for output. That way people can parse it and use it with jq, jmespath, etc.

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,253 @@
...
+        } catch (Exception e) {
+            logError("Unexpected error occurred: ", e);
+            return 1;
+        }
+    }
+
+    private Integer listOperationsForService() {
+        try {
+            Model model = assembleModel(modelPath);
+            System.out.println("Available Operations:\n" + model.getOperationShapes());
```
mtdowling, line 102: You could return each operation name separated by newlines instead. That's human readable, and still easily parsed and iterated over. Then skip the heading you're printing out.

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,253 @@
...
+                    break;
+                default:
+                    throw new IllegalArgumentException("Unsupported auth type: " + authType);
+            }
+        } else {
+            builder.authSchemeResolver(AuthSchemeResolver.NO_AUTH);
+        }
+    }
+
+    private void configureProtocol(DynamicClient.Builder builder, ShapeId serviceInput) {
```
mtdowling, line 200: It would be cool if a custom build of the CLI could register additional protocols and auth schemes.

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,253 @@
...
+        }
+    }
+
+    private Object executeClientCall(DynamicClient client) throws Exception {
+        if (input != null && inputPath != null) {
+            throw new IllegalArgumentException("Cannot specify both '--input-json' and '--input-path'. Please provide only one.");
+        }
+
+        if (input == null && inputPath == null) {
+            return client.call(operation).asObject();
```
mtdowling, line 236: Would return the document as-is here too so you can later serialize it as JSON

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,253 @@
...
+import software.amazon.smithy.java.client.protocols.rpcv2.RpcV2CborProtocol;
+import software.amazon.smithy.java.core.serde.document.Document;
+import software.amazon.smithy.java.dynamicclient.DynamicClient;
+import software.amazon.smithy.java.json.JsonCodec;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.shapes.ShapeId;
+
+@Command(name = "smithy-call", mixinStandardHelpOptions = true, version = "1.0",
+        description = "Smithy Java CLI")
+public class SmithyCall implements Callable<Integer> {
```
mtdowling, line 37: I'm fine with this not supporting pagination yet, but we should think about how it would work. On by default, or opt-in? How do we return multiple pages of results. Ideally we don't have to buffer them into memory since that's an OOM risk and it takes a while to start consuming results. Do we stream a JSON list? JSON lines?

mtdowling (changes requested):
Can you add a few basic tests?

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
franco2002lu: Agreed. I will redo the implementation using Logger

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
franco2002lu: I just talked to Brian about it. I think to start, we would want to copy what the aws CLI does and support [--max-items and --starting-token parameters](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-pagination.html#cli-usage-pagination-pagesize:~:text=How%20to%20use%20the%20%2D%2Dmax%2Ditems%20parameter). For more details I will look more into this and update the doc on how this will work.

franco2002lu:
Revision 2 here.

cli/build.gradle.kts
```diff
@@ -0,0 +1,103 @@
...
+    implementation("info.picocli:picocli:4.7.6")
+    annotationProcessor("info.picocli:picocli-codegen:4.7.6")
+
+    implementation("software.amazon.smithy:smithy-aws-traits:1.56.0")
```
rhernandez35, line 11: Use `implementation(libs.smithy.aws.traits)` and the like for the maven-sourced dependencies. This sources the version from [libs.version.toml](https://github.com/smithy-lang/smithy-java/blob/main/gradle/libs.versions.toml) so we have a single consistent set of dependencies for all modules. If you're editing this in an IDE, you should be able to get these with autocomplete.

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,297 @@
...
+    private String inputPath;
+
+    @Option(names = "--input-json", description = "JSON string containing input parameters for the operation")
+    private String inputJson;
+
+    @Option(names = "--url", description = "Endpoint URL for the service")
+    private String url;
+
+    @Option(names = { "-p", "--protocol" }, description = "Communication protocol to use (options: awsjson, rpcv2-cbor, restjson, and restxml)")
+    private String protocol;
```
rhernandez35, line 80: You can directly use your `ProtocolType` enum here and [picocli will automatically display its possible values and enable command-line completion for them](https://picocli.info/#_enum_types)

rhernandez35:
Looking good, only two minor comments from me.

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,288 @@
...
+import software.amazon.smithy.java.json.JsonCodec;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.ShapeId;
+
+@Command(name = "smithy-call", mixinStandardHelpOptions = true, version = "1.0",
+        description = "Smithy Call, send request to Smithy services using Smithy models")
+public final class SmithyCall implements Callable<Integer> {
+    private static final JsonCodec CODEC = JsonCodec.builder().build();
+    private static final Logger LOGGER = Logger.getLogger(SmithyCall.class.getName());
```
adwsingh, line 49: Using the InternalLogger is not feasible?

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,288 @@
...
+
+    private static final String[] BASE_RESOURCE_FILES = {
+            "aws.api.smithy",
+            "aws.auth.smithy",
+            "aws.customizations.smithy",
+            "aws.protocols.smithy"
+    };
+
+    @Parameters(index = "0", description = "Service Name (e.g. com.example#CoffeeShop)")
```
adwsingh, line 58: We should call this fully qualified service name or something.

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,288 @@
...
+    private String service;
+
+    @Parameters(index = "1", description = "Name of the operation to perform on the service", arity = "0..1")
```
adwsingh, line 61: This is required right? Why is the arity then `0..1` instead of 1. I think you can just skip specifying the arity.

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,288 @@
...
+                case REST_XML -> new RestXmlClientProtocol(serviceInput);
+            });
+        }
+    }
+
+    private void configureInputInterceptor(DynamicClient.Builder builder) {
+        if (inputJson != null || inputPath != null) {
+            builder.addInterceptor(new ClientInterceptor() {
+                @Override
+                public void readBeforeTransmit(RequestHook<?, ?, ?> hook) {}
```
adwsingh, line 262: What's the point of this?

franco2002lu: I originally left this here as an example of how we can configure interceptors for a custom-built version of the CLI. But for now this is not necessary so I will remove it.

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,288 @@
...
+import software.amazon.smithy.java.core.serde.document.Document;
+import software.amazon.smithy.java.dynamicclient.DynamicClient;
+import software.amazon.smithy.java.json.JsonCodec;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.ShapeId;
+
+@Command(name = "smithy-call", mixinStandardHelpOptions = true, version = "1.0",
+        description = "Smithy Call, send request to Smithy services using Smithy models")
+public final class SmithyCall implements Callable<Integer> {
```
adwsingh, line 47: If SmithyCallRunner is the intended entry point, do we need this class to be public?

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,288 @@
...
+
+        Document inputDocument;
+        if (inputJson != null) {
+            inputDocument = CODEC.createDeserializer(inputJson.getBytes(StandardCharsets.UTF_8)).readDocument();
+        } else {
+            String content = Files.readString(Path.of(inputPath), StandardCharsets.UTF_8);
+            inputDocument = CODEC.createDeserializer(content.getBytes(StandardCharsets.UTF_8)).readDocument();
+        }
+
+        return client.call(operation, inputDocument);
```
adwsingh, line 286: 1. You can avoid first converting to a string and then back to bytes when reading from a file.
2. You should close the deserializer.

```suggestion
        byte[] inputBytes;
        if (inputJson != null) {
            inputBytes = inputJson.getBytes(StandardCharsets.UTF_8);
        } else {
            inputBytes = Files.readAllBytes(Path.of(inputPath));
        }

        Document input;
        try (var d = CODEC.createDeserializer(inputBytes)) {
            input = d.readDocument();
        }

        return client.call(operation, input);
```

franco2002lu: I decided to keep `if (inputJson == null && inputPath == null) {
            return client.call(operation);
        }` to differentiate calls with input vs calls without input. Implemented rest of the changes.

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
franco2002lu: I don't see a reason as to why not, I will change this to package-private for now

adwsingh (approved):

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,288 @@
...
+    private static final Logger LOGGER = Logger.getLogger(SmithyCall.class.getName());
+
+    private static final String[] BASE_RESOURCE_FILES = {
+            "aws.api.smithy",
+            "aws.auth.smithy",
+            "aws.customizations.smithy",
+            "aws.protocols.smithy"
+    };
+
+    @Parameters(index = "0", description = "Fully qualified service Name (e.g. com.example#CoffeeShop)")
```
mtdowling, line 56: it would be nice if we could use a relative shape ID here in cases where it isn't ambiguous. So then you could do:

`smithy-call CoffeShop PutCoffee`

You see that the service is relative, so you look up all services and see if a single service has this name. If so, use it. Otherwise,  fail and list the possible services you found.

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,288 @@
...
+    private String service;
+
+    @Parameters(index = "1", description = "Name of the operation to perform on the service", arity = "0..1")
+    private String operation;
+
+    @Option(names = {"-v", "--verbose"}, description = "Enable verbose logging")
+    private boolean verbose;
+
+    @Option(names = { "-m", "--model-path" }, description = "Path to a directory containing all necessary Smithy models", required = true)
```
mtdowling, line 65: I see the value that captures this is a String[], implying it supports multiple model paths. That could be stated in the help docs. Do you know how passing multiple looks with CLI arguments?

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,288 @@
...
+    private String[] modelPath;
+
+    @Option(names = "--input-path", description = "Path to a JSON file containing input parameters for the operation")
+    private String inputPath;
+
+    @Option(names = "--input-json", description = "JSON string containing input parameters for the operation")
+    private String inputJson;
+
+    @Option(names = "--url", description = "Endpoint URL for the service")
```
mtdowling, line 74: The AWS CLI uses `--endpoint-url`. Should we copy that?

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,288 @@
...
+            "aws.protocols.smithy"
+    };
+
+    @Parameters(index = "0", description = "Fully qualified service Name (e.g. com.example#CoffeeShop)")
+    private String service;
+
+    @Parameters(index = "1", description = "Name of the operation to perform on the service", arity = "0..1")
+    private String operation;
+
+    @Option(names = {"-v", "--verbose"}, description = "Enable verbose logging")
```
mtdowling, line 62: The AWS CLI uses `--debug`. Would that be different than `--verbose` if we added support for that too?

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -0,0 +1,288 @@
...
+import software.amazon.smithy.java.client.rpcv2.RpcV2CborProtocol;
+import software.amazon.smithy.java.core.serde.ShapeSerializer;
+import software.amazon.smithy.java.core.serde.document.Document;
+import software.amazon.smithy.java.dynamicclient.DynamicClient;
+import software.amazon.smithy.java.json.JsonCodec;
+import software.amazon.smithy.model.Model;
+import software.amazon.smithy.model.shapes.OperationShape;
+import software.amazon.smithy.model.shapes.ShapeId;
+
+@Command(name = "smithy-call", mixinStandardHelpOptions = true, version = "1.0",
```
mtdowling, line 43: Is `--version` something that needs to be explicitly defined as a parameter?

franco2002lu: No, smithy-call --version works now

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
franco2002lu: It wouldn't be any different, I can have it support both --verbose and --debug

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
franco2002lu: yes, it would just be smithy-call -m [path1] -m [path2] .....

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
franco2002lu: I can have it support both --url and --endpoint-url

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
mtdowling: If it's the same, then `--debug` is probably better for consistency. If they're ultimately different though, I'd leave `--verbose` alone and don't claim `--debug` now.

franco2002lu: I think I will just change it to --debug, since I don't see a need of having both --verbose and --debug for the CLI and they would be the same ultimately. Do I use -d in this case too?

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
franco2002lu: That would be cool. I will have the CLI fail and list services when there are either 1) no matching relative shape ID, or 2) 2 or more matching relative shape IDs.

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
mtdowling: Nah, I'd pick one or the other. You can just do `--url`. The fewer options we support the better. It's less surface area for potentially breaking users if we have to make changes down the road.

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
mtdowling: Ok sounds good. No, let's not claim `-d` for this. We might want it for something else later.

mtdowling (changes requested):

mtdowling (approved):
LGTM. Let's squash and merge.

PR #621, opened 2025-03-28
Initial semi-consistent renaming of modules

haydenbaker:
*Description of changes:*
- Removes all sub-groups within the project, everything lives under a singular `software.amazon.smithy.smithy-java` group
- Renames several modules to follow a semi-consistent pattern - in general the pattern should be:
  -  prefix the module name with its parents
  -  modules under `plugin/` or `integration/` may
     - exclude plugin/integration from the name
     - reverse the order if it makes sense (`client-codegen` instead of `codegen-client`)
  - if a parent is a plural, the modules under it can do either order (`codecs/cbor-codec` or `codecs/codec-cbor`), just be consistent
 

mtdowling:
> waiters: it's just waiters... client-waiters? Bueller?

Sure, I could see this being client-waiters.

> mock-client-plugin: does it belong there?

I guess it could become `client-mock-plugin`.

> sigv4: needs a different home and name (maybe aws-auth-sigv4)

It's under aws, so how about aws-sigv4?

>dynamic-client: package namespace has dynamicclient in it (which is icky IMO), maybe dynamic.client and move it into a directory such as dynamic/dynamic-client alongside dynamic/dynamic-schemas (same deal with its package namespace)

I think this one may be too disruptive at the moment given all the stuff we're building with it right now. I think the DynamicClient going under client makes sense to an extent, and the root level dynamic-schemas is fine since it's not client or server.

mtdowling (approved):
@

PR #625, opened 2025-04-02
Add service schema to clients

mtdowling:
Clients now get access to the schema of the service, providing the shape ID and traits of the service allowed for code generation. This can be used to know what service is being called for telemetry, or implement things like endpoint resolvers or custom auth handling based on traits applied to a service.

Each generated ApiOperation now provides a getter called `service()` to get the generated ApiService for the client. Resources get this method too. The generated ApiService only provides the schema of the service. It doesn't provide a full accounting of every operation or resource within the service as there isn't a clear use for that, it adds complexity due to bidirectional references, and we want to avoid eagerly loading every shape in a service when the ApiService is loaded. This commit also simplifies the generated ApiResource for this same reason, removing the operation and sub-resource schemas.

DynamicClient will now automatically generate an ApiService based on the Smithy model.

----

You can get the service schema in an interceptor using:

```java
var client = ExampleClient.builder()
    .addInterceptor(new ClientInterceptor() {
        @Override
        public void readBeforeAttempt(RequestHook<?, ?, ?> hook) {
            var service = hook.operation().service();
        }
    })
    // ...
    .build();
```

Here is an example generated ApiService schema class.

```java
package software.amazon.smithy.java.examples.dynamodb.model;

import software.amazon.smithy.java.core.schema.ApiService;
import software.amazon.smithy.java.core.schema.Schema;
import software.amazon.smithy.model.shapes.ShapeId;
import software.amazon.smithy.model.traits.XmlNamespaceTrait;
import software.amazon.smithy.utils.SmithyGenerated;

/**
 * Service API schema.
 */
@SmithyGenerated
public final class DynamoDBApiService implements ApiService {
    private static final DynamoDBApiService $INSTANCE = new DynamoDBApiService();
    private static final Schema $SCHEMA = Schema.createService(ShapeId.from("com.amazonaws.dynamodb#DynamoDB_20120810"),
        XmlNamespaceTrait.builder().uri("http://dynamodb.amazonaws.com/doc/2012-08-10/").build());

    /**
     * Get an instance of this {@code ApiService}.
     *
     * @return An instance of this class.
     */
    public static DynamoDBApiService instance() {
        return $INSTANCE;
    }

    private DynamoDBApiService() {}

    @Override
    public Schema schema() {
        return $SCHEMA;
    }
}
```

Here is an example of an ApiOperation that references it:

```java
@SmithyGenerated
public final class DescribeTable implements ApiOperation<DescribeTableInput, DescribeTableOutput> {
    // ...

    @Override
    public DynamoDBApiService service() {
        return DynamoDBApiService.instance();
    }
}
```

Here's an example of an operation that is bound to a resource but still provides direct access to the service:

```java
@SmithyGenerated
public final class GetOrder implements ApiOperation<GetOrderInput, GetOrderOutput> {

    // ...

    @Override
    public CoffeeShopApiService service() {
        return CoffeeShopApiService.instance();
    }

    @Override
    public ApiResource boundResource() {
        return Order.instance();
    }
}
```

Here is an example of a generated resource that provides access to the service:

```java
/**
 * An Order resource, which has an id and describes an order by the type of coffee
 * and the order's status
 */
@SmithyGenerated
public final class Order implements ApiResource {
    private static final Order $INSTANCE = new Order();
    private static final Map<String, Schema> $IDENTIFIERS = Map.of("id", Schemas.UUID);
    private static final Map<String, Schema> $PROPERTIES = Map.of("coffeeType", CoffeeType.$SCHEMA,
        "status", OrderStatus.$SCHEMA);

    private static final Schema $SCHEMA = Schema.createResource(ShapeId.from("com.example#Order"));

    public static final ShapeId $ID = $SCHEMA.id();

    /**
     * Get an instance of this {@code ApiResource}.
     *
     * @return An instance of this class.
     */
    public static Order instance() {
        return $INSTANCE;
    }

    private Order() {}

    @Override
    public Schema schema() {
        return $SCHEMA;
    }

    @Override
    public Map<String, Schema> identifiers() {
        return $IDENTIFIERS;
    }

    @Override
    public Map<String, Schema> properties() {
        return $PROPERTIES;
    }

    @Override
    public CoffeeShopApiService service() {
        return CoffeeShopApiService.instance();
    }
}
```

A builder for a generated client hooks up the service schema in the config:

```java
final class Builder extends Client.Builder<TestServiceClient, Builder> implements TestSettings<Builder> {
        // ...

        private Builder() {
            configBuilder()
                .putSupportedAuthSchemes(testAuthSchemeSchemeFactory.createAuthScheme(testAuthSchemeScheme))
                .service(TestServiceApiService.instance());
        }

        // ...
```

### DIFF
--- a/codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/SymbolProperties.java
+++ b/codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/SymbolProperties.java
@@ -7,6 +7,7 @@ package software.amazon.smithy.java.codegen;
 
 import software.amazon.smithy.codegen.core.Property;
 import software.amazon.smithy.codegen.core.Symbol;
+import software.amazon.smithy.java.core.schema.ApiService;
 
 /**
  * Contains properties that may be added to symbols by smithy-java.
@@ -72,6 +73,13 @@ public final class SymbolProperties {
      * <p>This property is expected on all {@code Service} shape symbols.
      */
     public static final Property<Symbol> SERVICE_EXCEPTION = Property.named("service-exception");
+
+    /**
+     * Symbol representing the {@link ApiService} of a service.
+     *
+     * <p>This property is expected on all {@code Service} shape symbols.
+     */
+    public static final Property<Symbol> SERVICE_API_SERVICE = Property.named("service-api-service");
 
     private SymbolProperties() {}
 }
--- a/client/client-core/src/main/java/software/amazon/smithy/java/client/core/Client.java
+++ b/client/client-core/src/main/java/software/amazon/smithy/java/client/core/Client.java
@@ -47,14 +47,9 @@ public abstract class Client {
         }
 
         this.config = configBuilder.build();
-
         this.pipeline = ClientPipeline.of(config.protocol(), config.transport());
-
-        // TODO: Add an interceptor to throw service-specific exceptions (e.g., PersonDirectoryClientException).
         this.interceptor = ClientInterceptor.chain(config.interceptors());
-
         this.identityResolvers = IdentityResolvers.of(config.identityResolvers());
-
         this.typeRegistry = typeRegistry();
 
         if (config.retryStrategy() != null) {
@@ -184,7 +179,9 @@ public abstract class Client {
          */
         @SuppressWarnings("unchecked")
         public B withConfiguration(ClientConfig config) {
-            configBuilder.transport(config.transport())
+            configBuilder
+                    .service(config.service())
+                    .transport(config.transport())
                     .protocol(config.protocol())
                     .endpointResolver(config.endpointResolver())
                     .authSchemeResolver(config.authSchemeResolver())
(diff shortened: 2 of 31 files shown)


codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/SymbolProperties.java
```diff
@@ -73,5 +74,12 @@ public final class SymbolProperties {
      */
     public static final Property<Symbol> SERVICE_EXCEPTION = Property.named("service-exception");
 
+    /**
+     * Symbol representing the {@link ApiService} of a service.
+     *
+     * <p>This property is expected on all {@code Service} shape symbols.
+     */
+    public static final Property<Symbol> SERVICE_API_SERVICE = Property.named("service-api-servide");
```
adwsingh, line 82:
```suggestion
    public static final Property<Symbol> SERVICE_API_SERVICE = Property.named("service-api-service");
```

adwsingh (approved):

PR #629, opened 2025-04-03
Fix JsonDocuments equals implementation

rhernandez35:
The Document impls in JsonDocuments are records, so their default `equals` methods check that all the record members are equal. Since they all contain JsonSettings members, two JsonDocuments are only equal if they have the exact same JsonSettings instance.

### DIFF
--- a/codecs/json-codec/src/test/java/software/amazon/smithy/java/json/JsonDocumentTest.java
+++ b/codecs/json-codec/src/test/java/software/amazon/smithy/java/json/JsonDocumentTest.java
@@ -27,6 +27,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Consumer;
 import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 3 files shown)


adwsingh (approved):

mtdowling:
seems fine, just need to address some failing tests

codecs/json-codec/src/test/java/software/amazon/smithy/java/json/JsonDocumentTest.java
```diff
@@ -423,40 +422,6 @@ public static List<Arguments> checkEqualitySource() {
                 Arguments.of("{\"foo\":\"foo\"}", "{\"foo\":\"bar\"}", false));
     }
 
-    @Test
-    public void onlyEqualIfBothUseTimestampFormat() {
```
adwsingh, line 427: @mtdowling I think these tests are right. I don't think the equality of the documents should matter based on settings used to create them. As long as the deserialized contents are same, they should be same.

adwsingh: Spoke privately. This is a grey area, disabling tests for now.

PR #641, opened 2025-04-09
Add model bundle API

rhernandez35:
To be used by the MCP server bundle consumer. The AWS service bundler is ready to go but depends on another to-be-published PR.

### DIFF
--- a/model-bundler/bundle-api/model/bundle.smithy
+++ b/model-bundler/bundle-api/model/bundle.smithy
@@ -1,0 +1,14 @@
+$version: "2"
+
+namespace software.amazon.smithy.modelbundle.api
+
+structure Bundle {
+    @required
+    configType: String
+
+    @required
+    serviceName: String
+
+    @required
+    config: Document
+}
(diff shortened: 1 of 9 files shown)


adwsingh (approved):

model-bundler/bundle-api/model/bundle.smithy
```diff
@@ -0,0 +1,25 @@
...
+    configType: String
+
+    @required
+    serviceName: String
+
+    @required
+    config: Document
+}
+
+structure SimpleAwsConfig {
```
adwsingh, line 16: This should probably be outside this package, in an aws specific one.

rhernandez35: This doesn't need to be in here anyway, it's unused

adwsingh (approved):

rhernandez35:
this pr doesn't want to update with the latest revision for some reason, I'm going to close and open a new one

PR #646, opened 2025-04-11
Add Operation filtering

adwsingh:
*Issue #, if available:*

*Description of changes:*

Add ability to create a filtered service which only exposes certain operations of a service.


server/server-api/src/main/java/software/amazon/smithy/java/server/FilteredService.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * Returns the schema from the delegate service.
+     *
+     * @return The schema of the delegate service
+     */
+    @Override
+    public Schema schema() {
+        return delegate.schema();
```
rhernandez35, line 81: This schema will contain operations that are not returned by `getAllOperations()`. It's not a problem for now but I can see this violating some kind of assumption down the line.

rhernandez35 (approved):

PR #651, opened 2025-04-16
Produce service description bundles for AWS services

rhernandez35:
This commit adds the Bundler interface and an implementation for AWS services. The AWS service bundler will download the specified service's Smithy model from the public AWS model repository, extract its endpoints, remove some unnecessary elements from its model, and package it up into a Bundle. The smithy-mcp server will then load this bundle, parse it, and use the AwsAuthProvider defined in the aws-mcp-provider module to dynamically serve the bundle's model as a set of MCP tools and appropriately configure calls to that service.

This commit adds four new modules:

* :aws:aws-service-bundler: Pulls AWS models from the public model repo and turns them into AwsServiceMetadata files (defined in `aws-mcp-types`)
* :mcp:aws-mcp-provider: Consumes the service bundles produced by aws-service bundler and uses their contents to dynamically configure outbound requests in the smithy-mcp/smithy-call cli
* :mcp:aws-mcp-types: Smithy model definition for the AwsServiceMetadata and PreRequest structures. AwsServiceMetadata can live in its own module but I didn't want to create a fifth one.
* :model-bundler:bundle-cli: Runs the bundler

Next PR will be integrating `aws-mcp-provider` into the dynamic server. We also need to add real argument parsing to the CLI, because what I have in place right now is the bare minimum needed to get a functional tool.


aws/aws-service-bundler/build.gradle.kts
```diff
@@ -0,0 +1,18 @@
+plugins {
+    id("smithy-java.module-conventions")
+}
+
+description = "This module provides an adapter that allows Smithy Java to use auth from the AWS SDK for Java V2"
```
adwsingh, line 5: Need to fix the description.

aws/aws-service-bundler/src/main/java/software/amazon/smithy/java/aws/servicebundler/AwsServiceBundler.java
```diff
@@ -0,0 +1,201 @@
...
+                            .put(line.substring(0, line.indexOf("/")).toLowerCase(Locale.ROOT), line));
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private final ModelResolver resolver;
+    private final String serviceName;
+
+    AwsServiceBundler(String... args) {
```
adwsingh, line 55: Why are we taking a varargs?

model-bundler/bundle-api/src/main/java/software/amazon/smithy/modelbundle/api/ServiceLoaderLoader.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+package software.amazon.smithy.modelbundle.api;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.ServiceLoader;
+import java.util.function.Function;
+
+final class ServiceLoaderLoader {
```
adwsingh, line 13: 😆

settings.gradle.kts
```diff
@@ -85,6 +85,9 @@ include(":examples:mcp-server")
 include(":mcp")
 include(":mcp:mcp-schemas")
 include(":mcp:mcp-server")
-
+include(":mcp:aws-mcp-provider")
```
adwsingh, line 88: This should go under `:aws` instead of `:mcp`

adwsingh (approved):
Some minor feedback but overall LGTM.

settings.gradle.kts
rhernandez35: I'll move aws-mcp-types there too

aws/aws-service-bundler/src/main/java/software/amazon/smithy/java/aws/servicebundler/AwsServiceBundler.java
rhernandez35: As opposed to a `String[]`?

adwsingh: The constructor you are delegating to takes a single service name and you are just passing the first arg. Maybe I am missing something, but I don't understand why take multiple strings at all, since this class works only on a single service na,e.

rhernandez35: Because there are other providers that will take more arguments. I'm just passing the CLI args through with the first two arguments – output location and bundler type – redacted. Each bundle implementation is effectively a subcommand of the bundler CLI.

PR #655, opened 2025-04-17
Change generated Getters to have a 'get' prefix

adwsingh:
*Issue #, if available:*

*Description of changes:*

Adding the get prefix to method names to enable Kotlin property access syntax. 

### DIFF
--- a/codegen/codegen-core/src/it/java/software/amazon/smithy/java/codegen/test/ClientErrorCorrectionTest.java
+++ b/codegen/codegen-core/src/it/java/software/amazon/smithy/java/codegen/test/ClientErrorCorrectionTest.java
@@ -28,15 +28,15 @@ public class ClientErrorCorrectionTest {
                 .errorCorrection()
                 .build();
 
-        assertFalse(corrected.getBooleanMember());
+        assertFalse(corrected.isBoolean());
         assertEquals(corrected.getBigDecimal(), BigDecimal.ZERO);
         assertEquals(corrected.getBigInteger(), BigInteger.ZERO);
-        assertEquals(corrected.getByteMember(), (byte) 0);
-        assertEquals(corrected.getDoubleMember(), 0);
-        assertEquals(corrected.getFloatMember(), 0);
+        assertEquals(corrected.getByte(), (byte) 0);
+        assertEquals(corrected.getDouble(), 0);
+        assertEquals(corrected.getFloat(), 0);
         assertEquals(corrected.getInteger(), 0);
-        assertEquals(corrected.getLongMember(), 0);
-        assertEquals(corrected.getShortMember(), (short) 0);
+        assertEquals(corrected.getLong(), 0);
+        assertEquals(corrected.getShort(), (short) 0);
         assertEquals(corrected.getBlob(), ByteBuffer.allocate(0));
         assertEquals(corrected.getStreamingBlob().contentLength(), 0);
         corrected.getStreamingBlob().asByteBuffer().thenAccept(bytes -> assertEquals(0, bytes.remaining()));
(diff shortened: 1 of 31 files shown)


rhernandez35 (approved):

codegen/codegen-core/src/it/java/software/amazon/smithy/java/codegen/test/ClientErrorCorrectionTest.java
```diff
@@ -21,31 +21,32 @@
...
-        corrected.streamingBlob().asByteBuffer().thenAccept(bytes -> assertEquals(0, bytes.remaining()));
-        assertNull(corrected.document());
-        assertEquals(corrected.list(), List.of());
-        assertEquals(corrected.map(), Map.of());
-        assertEquals(corrected.timestamp(), Instant.EPOCH);
-        assertEquals(corrected.enumMember().type(), NestedEnum.Type.$UNKNOWN);
-        assertEquals(corrected.enumMember().value(), "");
-        assertEquals(corrected.intEnum().type(), NestedIntEnum.Type.$UNKNOWN);
-        assertEquals(corrected.intEnum().value(), 0);
+        assertFalse(corrected.getBooleanMember());
```
rhernandez35, line 31: This should just be `getBoolean`. No need for the `Member` suffix when we have the `get` prefix.

rhernandez35 (dismissed):

codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/CodegenUtils.java
```diff
@@ -255,6 +255,19 @@ public static String toDefaultValueName(String memberName) {
...
     }
 
+    /**
+     * Gets the name to use when defining the default value of a member.
+     *
+     * @param memberShape memberShape.
+     * @return Upper snake case name of default
+     */
+    public static String toGetterName(MemberShape memberShape, String memberName) {
+        var prefix = memberShape.isBooleanShape() ? "is" : "get";
```
rhernandez35, line 265: `memberShape.isBooleanShape()` is always false because it's a MemberShape. You need to see the shape type of its target.

codegen/codegen-core/src/it/java/software/amazon/smithy/java/codegen/test/ClientErrorCorrectionTest.java
mtdowling: Same with float, long, etc

codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/CodegenUtils.java
```diff
@@ -255,6 +255,19 @@ public static String toDefaultValueName(String memberName) {
         return CaseUtils.toSnakeCase(memberName).toUpperCase(Locale.ENGLISH) + "_DEFAULT";
     }
 
+    /**
+     * Gets the name to use when defining the default value of a member.
+     *
+     * @param memberShape memberShape.
+     * @return Upper snake case name of default
+     */
+    public static String toGetterName(MemberShape memberShape, String memberName) {
```
mtdowling, line 264: We should probably use `CodegenUtils.MEMBER_ESCAPER.escape` here like the old code did. If you pass the resolved name of getX to the escaper, it'll ensure we don't clash with anything like getClass.

mtdowling (changes requested):

codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/CodegenUtils.java
adwsingh: I created a new method escaper since there is no overlap with member escaper.

mtdowling (approved):

PR #669, opened 2025-04-23
Add ability to override call config in interceptor

mtdowling:
Interceptors now have a new hook, modifyBeforeCall that accepts the ApiOperation being called, input of the call, and ClientConfig. It is expected to return the same ClientConfig or an updated ClientConfig using ClientConfig#withRequestOverride.

This allows for much more dynamic calls with clients, like changing the protocol per/call, applying plugins per/call, and more.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/client/client-core/src/main/java/software/amazon/smithy/java/client/core/Client.java
+++ b/client/client-core/src/main/java/software/amazon/smithy/java/client/core/Client.java
@@ -14,6 +14,7 @@ import software.amazon.smithy.java.auth.api.identity.IdentityResolvers;
 import software.amazon.smithy.java.client.core.auth.scheme.AuthScheme;
 import software.amazon.smithy.java.client.core.auth.scheme.AuthSchemeResolver;
 import software.amazon.smithy.java.client.core.endpoint.EndpointResolver;
+import software.amazon.smithy.java.client.core.interceptors.CallHook;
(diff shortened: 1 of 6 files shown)


client/client-core/src/main/java/software/amazon/smithy/java/client/core/Client.java
```diff
@@ -75,20 +76,26 @@ protected <I extends SerializableStruct, O extends SerializableStruct> Completab
...
+
+        // Apply the given override before potentially applying interceptor based overrides.
+        var needsRebuild = false;
+        if (overrideConfig != null) {
+            needsRebuild = true;
+            callConfig = callConfig.withRequestOverride(overrideConfig);
+            callInterceptor = ClientInterceptor.chain(callConfig.interceptors());
+        }
+
+        var updatedConfig = callInterceptor.modifyBeforeCall(new CallHook<>(operation, callConfig, input));
```
rhernandez35, line 92: I think this should be applied before request overrides. We should let the most specific overrides dominate the configuration, and client-level settings are less specific than settings being applied on individual calls. I know I would be confused if overrides I passed to `Client::call` were silently ignored or overwritten.

rhernandez35 (approved):

PR #670, opened 2025-04-24
Improve smithy-call's handling of SigV4 authentication

franco2002lu:
The CLI now performs a cascade of checks for its SigV4 signing name in the following order:
1. First it checks if the shape has a `SigV4Trait` and uses its name if present
2. If not, it checks if the shape has a `ServiceTrait` and uses its ARN namespace if present
3. If neither trait is present, it defaults to the namespace of the shape ID


rhernandez35:
Can you start using feature branches instead of reusing the branch you initially developed on? There are dozens of unrelated commits associated with this PR.

rhernandez35 (approved):

chenliu0831:
Let's squash this for now - but yeah you should use new branch with a name related to the diff. Also in general the commit history when you first time publish PR should roughly be meaningful (I used to recommit the diff entirely for non-trivial changes)

PR #671, opened 2025-04-25
Add logging to display signing name used for sigv4 auth.

franco2002lu:
Added FINE level logging for SigV4 configuration.

### DIFF
--- a/cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
+++ b/cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
@@ -260,6 +260,7 @@ final class SmithyCall implements Callable<Integer> {
         if (auth != null) {
             switch (auth.authType.toLowerCase()) {
                 case "sigv4", "aws":
+                    LOGGER.fine("Configuring Sigv4 authentication");
                     if (auth.awsRegion == null) {
                         throw new IllegalArgumentException("SigV4 auth requires --aws-region to be set. Please provide the --aws-region option.");
                     }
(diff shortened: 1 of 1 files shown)


cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -260,6 +260,7 @@ private void configureAuth(DynamicClient.Builder builder, ShapeId serviceInput,
         if (auth != null) {
             switch (auth.authType.toLowerCase()) {
                 case "sigv4", "aws":
+                    LOGGER.fine("Configuring Sigv4 authentication");
```
rhernandez35, line 263: Is this log message helpful? The presence of the log on line 273 should be enough to know that SigV4 is in use.

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
```diff
@@ -269,6 +270,7 @@ private void configureAuth(DynamicClient.Builder builder, ShapeId serviceInput,
                                     .map(ServiceTrait::getArnNamespace)
                                     .orElse(serviceInput.getNamespace().toLowerCase()));
 
+                    LOGGER.fine("Using service signing name detected in model: " + signingName);
```
rhernandez35, line 273: Can we move to using the smithy-logger interface in a follow-up PR instead of using JUL directly? We want to use format strings like `log("Using signing service name {}", signingName);` because they prevent the creation of unnecessary garbage in the form of concatenated strings. JUL does not support that.

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
franco2002lu: I can remove this line and combined it into other log.

cli/src/main/java/software/amazon/smithy/java/cli/SmithyCall.java
franco2002lu: by the smithy-logger interface, do you mean InternalLogger or is that something different?

rhernandez35 (approved):

PR #689, opened 2025-05-14
Shade Jackson

adwsingh:
*Issue #, if available:*

*Description of changes:*

Shade Jackson so that it doesn't interfere with customers jackson dependencies.


sugmanue (approved):

buildSrc/src/main/kotlin/smithy-java.publishing-conventions.gradle.kts
```diff
@@ -24,8 +32,7 @@ publishing {
     publications {
         afterEvaluate {
             create<MavenPublication>("mavenJava") {
-                from(components["java"])
-
+                from(extension.customComponent ?: components["java"])
```
haydenbaker, line 35: I don't foresee needing the configurability of the component to be published, so:
```
            if (components.names.contains("shadow")) {
                // Use the shadow component if its exists
                from(components["shadow"])
            } else {
                // Otherwise, use the standard java component
                from(components["java"])
            }
```
should suffice

codecs/json-codec/build.gradle.kts
```diff
@@ -11,3 +12,43 @@ dependencies {
...
+    )
+}
+
+artifacts {
+    configurations.archives
+        .get()
+        .artifacts
+        .clear()
+    archives(tasks.shadowJar.get())
+}
```
haydenbaker, line 54: Is this really needed?

codecs/json-codec/build.gradle.kts
```diff
@@ -11,3 +12,43 @@ dependencies {
...
+
+tasks.jar {
+    enabled = true
+    dependsOn(tasks.shadowJar)
+    outputs.files(
+        tasks.shadowJar
+            .get()
+            .outputs.files,
+    )
+}
```
haydenbaker, line 46:
```suggestion
    implementation(libs.jackson.core)
    shadow(libs.jackson.core)
}

tasks {
    shadowJar {
        archiveClassifier.set("")
        mergeServiceFiles()

        dependencies {
            relocate("com.fasterxml.jackson.core", "software.amazon.smithy.java.shaded.com.fasterxml.jackson.core")
            exclude(project(":core"))
        }
    }
    jar {
        finalizedBy(shadowJar)
    }
}

```

haydenbaker: Maybe I ran it wrong, but I don't think it was actually shading in jackson.

haydenbaker (changes requested):
Maybe I ran it wrong, but I don't think it was actually shading in jackson, see my comments.

buildSrc/src/main/kotlin/smithy-java.publishing-conventions.gradle.kts
adwsingh: We do. We use the shadow plugin in the CLIs too and we don't want them published to Maven.

haydenbaker: What I'm asking is if we will ever have more than `"java"` and `"shadow"`

haydenbaker (approved):

PR #719, opened 2025-05-28
Make jars reproducible

adwsingh:
*Issue #, if available:*

*Description of changes:*

This makes jars reproducible to an extent. Spot checked MD5 sum of some jars remain same across clean builds and on different machines.


rhernandez35 (approved):

yasmewad (approved):
LGTM! Curious though does this have any effect on the build time at all? I assume it is negligible?

PR #721, opened 2025-05-28
Resolve AWS services in the AWS bundler by sigv4 signing name

rhernandez35:
Now, we input the sigv4 signing name to the bundler and use that to produce the bundles. This doesn't _exactly_ align with the experience we want but it's closer than we were before.


shwdean (approved):
LGTM

PR #731, opened 2025-05-31
Add installed tag to already installed servers

adwsingh:
*Issue #, if available:*

*Description of changes:*


rhernandez35 (approved):
Maybe we should filter installed by default and only show them on request

adwsingh:
@rhernandez35 sounds good. Will do a follow up.

PR #740, opened 2025-06-02
Add env to Mcp server config

adwsingh:
*Issue #, if available:*

*Description of changes:*

So that we don't clobber existing client configs.

### DIFF
--- a/mcp/mcp-cli-api/model/main.smithy
+++ b/mcp/mcp-cli-api/model/main.smithy
@@ -95,8 +95,14 @@ map McpServerConfigs {
 structure McpServerConfig {
     command: String
     args: ArgsList
+    env: EnvVars
 }
 
 list ArgsList {
     member: String
+}
+
+map EnvVars {
+    key: String
+    value: String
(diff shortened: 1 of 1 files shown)


jlhood:
Thanks for fixing this. However, is there a longer term fix so you don't have to keep updating this anytime the structure of mcp.json changes? For example, Q CLI also supports [a timeout property](https://github.com/aws/amazon-q-developer-cli/blob/3311ed237c6ebe4ab5ea5edf319bc88d445d2f76/crates/cli/src/cli/chat/tools/custom_tool.rs#L44) that needs to be added here as well. Q CLI can add other properties in the future which puts you on a treadmill trying to keep up. Other tools may support their own custom properties, e.g., specifying tool filtering or tool auto-approve settings. It would be better if you consume the mcp.json as raw JSON and then don't touch any entries other than the one you're modifying.

rhernandez35:
Yes, there's a simple change we can make to avoid losing custom configs. I'll put it on the backlog.

PR #755, opened 2025-06-06
Cache dependencies in CI runs

rhernandez35:
https://github.com/actions/setup-java?tab=readme-ov-file#caching-packages-dependencies

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,7 @@ jobs:
         with:
           java-version: ${{ matrix.java }}
           distribution: 'corretto'
+          cache: 'gradle'
 
       - name: Cache compiled buildscripts
         uses: actions/cache@v4


adwsingh:
The setup gradle action we use explicitly asks to avoid using this, https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#incompatibility-with-other-caching-mechanisms

adwsingh:
Where do you see that? In the last PR I can see it restoring 600MB https://github.com/smithy-lang/smithy-java/actions/runs/15481667001?pr=756

rhernandez35:
> Where do you see that? In the last PR I can see it restoring 600MB https://github.com/smithy-lang/smithy-java/actions/runs/15481667001?pr=756

Huh, you're right. I was looking at the wrong action. It seems like we're still not caching _something_ relevant though, since [this run from my PR](https://github.com/smithy-lang/smithy-java/actions/runs/15481593352/job/43588610816?pr=755) is still over a minute faster than one from HEAD. Maybe I just got lucky and got a fast host.

Issue #737, opened 2025-06-02
CORS Support Not Working Properly in Generated Netty Server


dpacheco-groupieticket:
**Description**:  
The generated Netty server from `smithy-java-server-codegen` doesn't properly handle CORS requests despite having `@cors` traits in the Smithy model. While OPTIONS preflight requests work, actual requests (POST/GET) don't include CORS headers in responses.

**Current Behavior**:  
1. OPTIONS preflight requests return correct CORS headers  
2. Actual requests (POST/GET) return:  
   - Correct status codes (200/201)  
   - Correct response bodies  
   - But missing `Access-Control-Allow-Origin` and other CORS headers  
3. Browser shows error: "CORS header 'Access-Control-Allow-Origin' missing"

**Expected Behavior**:  
All responses should include CORS headers when:  
1. The request contains an `Origin` header  
2. The origin matches the `@cors` trait configuration  
3. For both preflight (OPTIONS) and actual requests  

**Reproduction Steps**:  
1. Model with `@cors` trait:  
```smithy
@cors(origin: "http://localhost:5173")
service MyService {
    operations: [MyOperation]
}
```  
2. Generate Java server code  
3. Start server and make CORS request from browser  
4. Observe missing headers in actual requests

rhernandez35:
Merged in #756, thank you for the contribution.

PR #756, opened 2025-06-06
Adding Cors handler to http request handler in netty server

dpacheco-groupieticket:
*Issue https://github.com/smithy-lang/smithy-java/issues/737*

###  Netty service
This PR contains a new handler to respond with require headers when [smithy.api#cors](https://smithy.io/2.0/spec/http-bindings.html#smithy-api-cors-trait) trait is set in the service. 

The CorsHandler supports the case of having multiple origins, when the origin value contains multiple origins separated by commas it will check against the origin in attached in the request. 

### Json Serializer Test fix.

The  `testPrettyPrinting()` test fails due to differences in line endings between operating systems. Windows uses \r\n (CRLF), while Unix-based systems (Linux/macOS) use \n (LF). The expected string, defined in a Java text block, contains Unix-style line endings (\n), but the actual output (result) contains Windows-style line endings (\r\n), causing the assertion to mismatch despite identical content. To ensure consistent behavior across platforms, we normalize line endings by removing \r before comparison or use System.lineSeparator() for OS-agnostic formatting.

## Test

No unit tests are configure for the netty-server subpackage. I manually tested from an external project and the requests returns the expected headers.


server/server-core/src/main/java/software/amazon/smithy/java/server/core/CorsHeaders.java
```diff
@@ -0,0 +1,80 @@
...
+            return false;
+        }
+
+        return job.request() != null
+                && job.request().headers() != null
+                && job.request().headers().hasHeader("origin");
+    }
+
+    private static Optional<String> getConfiguredOrigin(HttpJob job) {
+        return Optional.ofNullable(job.operation()
```
rhernandez35, line 59: `Optional.ofNullable` is not a replacement for the venerable null-check. `Optional` is best used when crossing API boundaries; `private Optional<T>` method is an anti-pattern. This would be best written as

```
var corsTrait = job.operation().getApiOperation()
  .service()
  .schema()
  .getTrait(CORS_TRAIT);
if (corsTrait != null) {
  return corsTrait.getOrigin();
}
return null;
```

especially since the sole caller of this method immediately calls `orElse(null)` on its return value.

server/server-core/src/main/java/software/amazon/smithy/java/server/core/CorsHeaders.java
```diff
@@ -0,0 +1,80 @@
...
+                job.operation().getApiOperation().service() == null
+                ||
+                job.operation().getApiOperation().service().schema() == null
+                ||
+                !job.operation().getApiOperation().service().schema().hasTrait(CORS_TRAIT)) {
+            return false;
+        }
+
+        return job.request() != null
+                && job.request().headers() != null
```
rhernandez35, line 54: These will never be null

server/server-core/src/main/java/software/amazon/smithy/java/server/core/CorsHeaders.java
```diff
@@ -0,0 +1,80 @@
...
+        headers.put("Access-Control-Max-Age", List.of("600"));
+
+        return headers;
+    }
+
+    private static boolean shouldAddCorsHeaders(HttpJob job) {
+        if (job == null || job.operation() == null
+                || job.operation().getApiOperation() == null
+                ||
+                job.operation().getApiOperation().service() == null
```
rhernandez35, line 45: Did the code formatter make you split the lines up like this?

server/server-core/src/main/java/software/amazon/smithy/java/server/core/CorsHeaders.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        String requestOrigin = job.request().headers().firstValue("origin");
+        Optional<String> configuredOrigin = getConfiguredOrigin(job);
+
+        if (!isOriginAllowed(configuredOrigin.orElse(null), requestOrigin)) {
+            return Map.of();
+        }
+
+        Map<String, List<String>> headers = new HashMap<>();
```
rhernandez35, line 31: We're allocating and populating a HashMap only to immediately add its contents to another map. We can avoid the unnecessary allocation by mutating the response headers directly in this method.

server/server-core/src/main/java/software/amazon/smithy/java/server/core/CorsHeaders.java
```diff
@@ -0,0 +1,80 @@
...
+        headers.put("Access-Control-Allow-Origin", List.of(requestOrigin));
+        headers.put("Access-Control-Allow-Methods", List.of("GET, POST, PUT, DELETE, OPTIONS"));
+        headers.put("Access-Control-Allow-Headers",
+                List.of("*,Access-Control-Allow-Headers,Access-Control-Allow-Methods,Access-Control-Allow-Origin,Amz-Sdk-Invocation-Id,Amz-Sdk-Request,Authorization,Content-Length,Content-Type,X-Amz-User-Agent,X-Amzn-Trace-Id"));
+        headers.put("Access-Control-Max-Age", List.of("600"));
```
rhernandez35, line 36: These lists can be statically created and reused for all jobs

server/server-core/src/main/java/software/amazon/smithy/java/server/core/CorsHeaders.java
```diff
@@ -0,0 +1,80 @@
...
+package software.amazon.smithy.java.server.core;
+
+import static software.amazon.smithy.java.core.schema.TraitKey.CORS_TRAIT;
+
+import java.util.*;
+import software.amazon.smithy.java.logging.InternalLogger;
+import software.amazon.smithy.model.traits.CorsTrait;
+
+public final class CorsHeaders {
+    private static final InternalLogger LOGGER = InternalLogger.getLogger(CorsHeaders.class);
```
rhernandez35, line 15: Unused

server/server-netty/src/main/java/software/amazon/smithy/java/server/netty/HttpRequestHandler.java
```diff
@@ -98,6 +104,7 @@ private void writeResponse(Channel channel, HttpJob job) {
             if (serializedValue.contentType() != null) {
                 response.headers().set("content-type", serializedValue.contentType());
             }
+            CorsHeaders.of(job).forEach(response.headers()::set);
```
rhernandez35, line 107: `response.headers().putHeaders(CorsHeaders.of(job));` would be better. The current method requires allocating and invoking a state-capturing lambda, which is surprisingly hard for the JIT to inline and optimize away.

The best option, though, is to set the headers in `CorsHeaders#of`.

server/server-core/src/main/java/software/amazon/smithy/java/server/core/CorsHeaders.java
```diff
@@ -0,0 +1,80 @@
...
+
+package software.amazon.smithy.java.server.core;
+
+import static software.amazon.smithy.java.core.schema.TraitKey.CORS_TRAIT;
+
+import java.util.*;
+import software.amazon.smithy.java.logging.InternalLogger;
+import software.amazon.smithy.model.traits.CorsTrait;
+
+public final class CorsHeaders {
```
rhernandez35, line 14: Can you write unit tests for this class?

rhernandez35:
Thank you for the PR!

server/server-core/src/main/java/software/amazon/smithy/java/server/core/CorsHeaders.java
dpacheco-groupieticket: I ran the `spotlessApply` task but the format still the same. I will format it manually in a better way. I am wondering which of this properties can be null at this point? `job.operation().getApiOperation().service().schema()`

server/server-core/src/main/java/software/amazon/smithy/java/server/core/CorsHeaders.java
```diff
@@ -0,0 +1,76 @@
...
+                ||
+                !job.operation().getApiOperation().service().schema().hasTrait(CORS_TRAIT)) {
+            return false;
+        }
+        return job.request().headers().hasHeader("origin");
+    }
+
+    private static String getConfiguredOrigin(HttpJob job) {
+        var corsTrait = job.operation().getApiOperation()
+                .service()
```
rhernandez35, line 54:
```suggestion
        var corsTrait = job.operation()
                .getApiOperation()
                .service()
```

PR #760, opened 2025-06-10
Print human-readable titles alongside MCP server identifiers in list subcommand

rhernandez35:
We now print a friendly title with each server. The server identifier is bolded so users will know to use it with `start-server` and `install`. We can move the title into the body of the listing if we think we'll need additional clarity.

### DIFF
--- a/mcp/mcp-bundle-api/src/main/java/software/amazon/smithy/mcp/bundle/api/Registry.java
+++ b/mcp/mcp-bundle-api/src/main/java/software/amazon/smithy/mcp/bundle/api/Registry.java
@@ -11,9 +11,35 @@ import software.amazon.smithy.mcp.bundle.api.model.BundleMetadata;
 
 public interface Registry {
 
+    final class RegistryEntry {
+        private final String title;
+        private final BundleMetadata bundleMetadata;
+
+        public RegistryEntry(String title, BundleMetadata bundleMetadata) {
+            this.title = title;
+            this.bundleMetadata = bundleMetadata;
+        }
+
+        /**
+         * Provides a human-readable title for the bundle for display purposes.
+         *
+         * @return the human-facing title of the registry entry
+         */
+        public String getTitle() {
+            return title;
+        }
+
+        /**
+         * @return the Bundle metadata
+         */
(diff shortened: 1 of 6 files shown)


mcp/mcp-bundle-api/src/main/java/software/amazon/smithy/mcp/bundle/api/Registry.java
```diff
@@ -11,9 +11,35 @@
 
 public interface Registry {
 
+    final class RegistryEntry {
```
rhernandez35, line 14: This is intentionally not a record. I don't want this to be usable in record destructuring statements.

mcp/mcp-cli/src/main/java/software/amazon/smithy/java/mcp/cli/commands/ListBundles.java
```diff
@@ -29,13 +29,14 @@ public class ListBundles extends SmithyMcpCommand {
...
-                .listMcpBundles()) {
+        for (Registry.RegistryEntry entry : registry.listMcpBundles()) {
+            var bundle = entry.getBundleMetadata();
             boolean isInstalled = installedBundles.contains(bundle.getName());
             var commandLine = spec.commandLine();
             System.out.println(commandLine
                     .getColorScheme()
-                    .string("@|bold " + bundle.getName() + (isInstalled ? " [installed]" : "") + "|@"));
+                    .string("@|bold " + bundle.getName() + (isInstalled ? " [installed]" : "") + "|@: "
+                            + entry.getTitle()));
```
adwsingh, line 39: Can you add a println at the end of description. Otherwise descriptions are clobberred.

adwsingh (approved):

mcp/mcp-cli/src/main/java/software/amazon/smithy/java/mcp/cli/commands/ListBundles.java
rhernandez35: What do you mean by "clobbered?" The output appears fine on my screen, if not a bit cramped.

PR #763, opened 2025-06-10
Fix aws bundle allowlisting

adwsingh:
*Issue #, if available:*

*Description of changes:*

The current bundling mechanism doesn't respect if allowedApis/blockedApis are provided. Changing to default to only readOnly if no allowlist is provided.


aws/aws-mcp-cli-commands/src/main/java/software/amazon/smithy/java/aws/mcp/cli/commands/AddAwsServiceBundle.java
```diff
@@ -38,16 +37,23 @@ public class AddAwsServiceBundle extends AbstractAddBundle {
     @Option(names = "--include-write-apis",
             description = "Include write APIs in the MCP server",
             defaultValue = "false")
-    boolean includeWriteApis;
+    Boolean includeWriteApis;
```
rhernandez35, line 40: This won't default to `Boolean.FALSE`?

rhernandez35 (approved):

aws/aws-mcp-cli-commands/src/main/java/software/amazon/smithy/java/aws/mcp/cli/commands/AddAwsServiceBundle.java
adwsingh: My bad, missed adding that to the commit.

aws/aws-mcp-cli-commands/src/main/java/software/amazon/smithy/java/aws/mcp/cli/commands/AddAwsServiceBundle.java
```diff
@@ -36,18 +35,24 @@ public class AddAwsServiceBundle extends AbstractAddBundle {
...
         var bundleBuilder = AwsServiceBundler.builder()
-                .serviceName(awsServiceName)
-                .exposedOperations(allowedTools())
-                .blockedOperations(blockedTools());
-        if (!includeWriteApis) {
-            bundleBuilder = bundleBuilder.readOnlyOperations();
+                .serviceName(awsServiceName);
+        //If nothing is specified then default to only readOnlyOperations.
+        if (allowedApis != null) {
+            // User explicitly specified allowed APIs - use only those
```
shwdean, line 47: Sorry just noticed this PR. I had previously designed the feature so that you could easily additionally include a small number of write APIs (or read APIs that don't follow the common naming scheme e.g. 'Query' and 'Scan' in DDB) without having to re-specify the entire list of read-only operations. With this change, you you need to individually allow every read API if you also want to use a single write API.

adwsingh: So the existing code didn't respect a users allowedApis flag, unless they explicitly pass a `--include-write-apis=false` . And if they don't pass it it automatically includes all APIs irrespective of the allowlist. TBH I think includeWriteApis is a poor choice of input. We should instead have had `--readOnlyApis` which adds all readOnlyApis and then `--allowed` lets you add more APIs and we can assume `--readOnlyApis` to be true by default if nothing is passed.

PR #772, opened 2025-06-11
Allow read only operations by default

shwdean:
Allows users to incrementally allow write APIs without re-specifying the entire list of read-only APIs


rhernandez35 (approved):

aws/aws-mcp-cli-commands/src/main/java/software/amazon/smithy/java/aws/mcp/cli/commands/AddAwsServiceBundle.java
```diff
@@ -34,24 +34,25 @@ public class AddAwsServiceBundle extends AbstractAddBundle {
...
 
     @Override
     protected CliBundle getNewToolConfig() {
         var bundleBuilder = AwsServiceBundler.builder()
                 .serviceName(awsServiceName);
-        //If nothing is specified then default to only readOnlyOperations.
+        if (readOnlyApis) {
+            // Include read-only apis unless disabled
+            bundleBuilder.readOnlyOperations();
+        }
```
adwsingh, line 49: I think this should be `Boolean` and we should still only respect the allowedApis if explicitly passed. Right now there is no way for a customer to only create a bundle for a single API without blockingListing every other one.

IMO what would be better is this,

1. If nothing is passed default to read-only-apis.
2. If an allow list is passed use only that. Users can optionally also pass read-only-apis and then they get the union of allowList and readOnlyApis.

PR #787, opened 2025-06-18
Handle additionalInput without modifying the actual input model

adwsingh:
*Issue #, if available:*

*Description of changes:*

Modifying the actual input model has other side affects like affecting RestJson serialization. To avoid these we now create proxy operations which delegate to the actual operations.

### DIFF
--- a/client/client-core/src/main/java/software/amazon/smithy/java/client/core/Client.java
+++ b/client/client-core/src/main/java/software/amazon/smithy/java/client/core/Client.java
@@ -80,11 +80,20 @@ public abstract class Client {
         IdentityResolvers callIdentityResolvers = identityResolvers;
         ClientInterceptor callInterceptor = interceptor;
 
-        // First apply overrides from interceptors.
-        ClientConfig callConfig = callInterceptor.modifyBeforeCall(new CallHook<>(operation, config, input));
-        // Overrides given per/operation take precedence over interceptors.
+        //If there is an override config first apply that before sending to interceptors.
+        ClientConfig callConfig = config;
         if (overrideConfig != null) {
             callConfig = callConfig.withRequestOverride(overrideConfig);
+        }
+        ClientConfig afterInterceptionConfig =
+                callInterceptor.modifyBeforeCall(new CallHook<>(operation, callConfig, input));
+        if (afterInterceptionConfig != null && afterInterceptionConfig != callConfig) {
+            if (overrideConfig != null) {
+                callConfig = afterInterceptionConfig.withRequestOverride(overrideConfig);
+            } else {
+                callConfig = afterInterceptionConfig;
+            }
+
         }
 
         // Rebuild the pipeline, resolvers, etc if the config changed.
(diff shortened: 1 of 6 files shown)


client/client-core/src/main/java/software/amazon/smithy/java/client/core/Client.java
```diff
@@ -80,12 +80,16 @@ protected <I extends SerializableStruct, O extends SerializableStruct> Completab
         IdentityResolvers callIdentityResolvers = identityResolvers;
         ClientInterceptor callInterceptor = interceptor;
 
-        // First apply overrides from interceptors.
-        ClientConfig callConfig = callInterceptor.modifyBeforeCall(new CallHook<>(operation, config, input));
-        // Overrides given per/operation take precedence over interceptors.
+        //If there is an override config first apply that before sending to interceptors.
+        ClientConfig callConfig = config;
```
rhernandez35, line 84: @mtdowling and I discussed this on the original CR. I thought it would be confusing if request overrides specified at an individual call site could be trumped by global overrides, so override configs should have highest precedence. Why does it need to be changed?

adwsingh: Modified to now give the interceptor a view of the overriden config but also again override the config.

server/server-proxy/src/main/java/software/amazon/smithy/java/server/ProxyService.java
```diff
@@ -81,31 +84,82 @@ private ProxyService(Builder builder) {
...
+                            true,
                             model,
+                            service);
+                    addOperation(serverOperation, operationName);
+                }
+            }
+        }
+
+        for (var operation : containedOperations) {
+            if (!operation.hasTrait(ProxyOperationTrait.class)) {
```
rhernandez35, line 110: this can be an `else` in the first iteration over containedOperations

model-bundle/model-bundle-api/src/main/java/software/amazon/smithy/modelbundle/api/ModelBundles.java
```diff
@@ -50,59 +54,84 @@ private static Model getModel(SmithyBundle bundle) {
...
-                        .input(additionalInputShape)
-                        .build());
-            } else {
-                var shape = model.expectShape(op.getInputShape(), StructureShape.class);
-                for (var member : shape.members()) {
-                    if (model.expectShape(member.getTarget()).hasTrait(StreamingTrait.class)) {
-                        b.removeShape(op.toShapeId());
-                        skipOperation = true;
-                        break;
+        for (var service : model.getServiceShapes()) {
```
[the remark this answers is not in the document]
adwsingh: Yeah I missed this is the SmithyBundle we are working on here.

server/server-proxy/src/main/java/software/amazon/smithy/java/server/ProxyService.java
adwsingh: Refactored code and added a comment to clarify why we are doing it this way.

rhernandez35 (approved):

PR #804, opened 2025-06-30
Add create command and have a concept of local bundles

adwsingh:
*Issue #, if available:*

*Description of changes:*

  - Add create command infrastructure with pluggable bundle creation support
  - Implement local bundle concept to distinguish between registry and locally created bundles
  - Add AWS service bundle creation via create aws-service-mcp subcommand
  - Add generic MCP bundle creation via create generic-mcp subcommand

### DIFF
--- a/aws/aws-mcp-cli-commands/src/main/java/software/amazon/smithy/java/aws/mcp/cli/commands/CreateAwsServiceBundle.java
+++ b/aws/aws-mcp-cli-commands/src/main/java/software/amazon/smithy/java/aws/mcp/cli/commands/CreateAwsServiceBundle.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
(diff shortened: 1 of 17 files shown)


aws/aws-mcp-cli-commands/src/main/java/software/amazon/smithy/java/aws/mcp/cli/commands/CreateAwsServiceBundle.java
```diff
@@ -0,0 +1,89 @@
...
+        String awsServiceName;
+
+        @Option(names = {"-a", "--allowed-apis"}, description = "List of APIs to expose in the MCP server")
+        protected Set<String> allowedApis;
+
+        @Option(names = {"-b", "--blocked-apis"}, description = "List of APIs to hide in the MCP server")
+        protected Set<String> blockedApis;
+
+        @Option(names = "--read-only-apis",
+                description = "Include read only APIs in the MCP server")
```
brnstz, line 86: I assume this means excluding write APIs? It could use another "only". `Include only read-only APIs`?

PR #809, opened 2025-07-08
Make SchemaGuidedDocumentBuilder public

shduke:
This will allow consumers to more easily create ShapeBuilders, which are use as inputs to other Smithy APIs.


rhernandez35:
You should access this through [SchemaConverter#createDocumentBuilder](https://github.com/smithy-lang/smithy-java/blob/9230558b7e15478ff18ad6a19fb8ca1262343bff/dynamic-schemas/src/main/java/software/amazon/smithy/java/dynamicschemas/SchemaConverter.java#L59).

shduke:
That actually works better thank you. I will close this PR

PR #825, opened 2025-07-25
Add Prompts support to the MCP Server

yasmewad:
**NOTE: The main goal with the PR is to get feedback and discuss around ways to ingest the model / prompts appropriately**.

This adds prompt support to the MCP server, letting LLMs get helpful context directly from Smithy models.

### What's New

Core Features:
• Define prompts in Smithy using @prompts trait
• MCP server automatically exposes them via `prompts/list` and `prompts/get`
• Template support with `{{placeholder}}` syntax for dynamic content.

Key Components:
• `PromptLoader` - discovers prompts from Smithy models
• `PromptProcessor` - handles template substitution and validation

### Testing

- Added unit tests for various cases of PromptLoader, PromptProcessor and the MCP Server itself.
- Added example in ProxyMcpServerExample and ran it with Q CLI. 
- Will add more testing examples in next revision after discussion has settled. 

#### This is what `/prompts` in Q CLI which correctly demonstrates the data from the prompts trait. Will add more calling examples later.

Screenshot 2025-07-24 at 4 24 45 PM

#### User experience for `prompts/get` or `@prompts` invocation in Q CLI.
Screenshot 2025-07-24 at 6 21 09 PM


mcp/mcp-schemas/model/main.smithy
```diff
@@ -182,3 +182,68 @@ structure TextContent {
 
     text: String
 }
+
+structure ListPromptsResult {
```
yasmewad, line 186: I will not merge this as is. I plan to add more documentation around what these fields represent.

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpServer.java
```diff
@@ -113,12 +120,29 @@ private void handleRequest(JsonRpcRequest req) {
...
                                 .build());
+                case "prompts/list" -> writeResponse(req.getId(),
+                        ListPromptsResult.builder().prompts(prompts.values().stream().toList()).build());
+                case "prompts/get" -> {
+                    var promptName = req.getParams().getMember("name").asString();
+                    var promptArguments = req.getParams().getMember("arguments");
+
+                    var prompt = prompts.get(promptName);
+
+                    if (prompt == null) {
```
yasmewad, line 138: Should we respond with PromptMessage too ? Like we do in the file PromptProcessor.java?

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/PromptLoader.java
```diff
@@ -0,0 +1,113 @@
...
+
+                    promptInfos.put(
+                            promptName,
+                            PromptInfo
+                                    .builder()
+                                    .name(promptName)
+                                    .description(promptTemplateDefinition.getDescription())
+                                    .template(
+                                            promptTemplateDefinition.getPreferWhen().isPresent()
+                                                    ? templateString + TOOL_PREFERENCE_PREFIX
```
yasmewad, line 61: Any suggestions on making this concatenation even more efficient yet elegant?

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpServerBuilder.java
```diff
@@ -22,6 +23,7 @@ public final class McpServerBuilder {
     OutputStream os;
     List<Service> serviceList = new ArrayList<>();
     List<McpServerProxy> proxyList = new ArrayList<>();
+    List<Model> modelList = new ArrayList<>();
```
yasmewad, line 26: This is the main point I want feedback. Where and how should we provide the entry point of prompts.

mcp/mcp-schemas/model/main.smithy
yasmewad: Essentially, though they mirror the MCP Spec and are consistent with naming of things used for `ToolInfo` class.

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/PromptProcessor.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * Applies template arguments to a template string with maximum efficiency.
+     *
+     * @param template The template string containing {{placeholder}} patterns
+     * @param arguments Document containing replacement values
+     * @return The template with all placeholders replaced
+     */
+    public String applyTemplateArguments(String template, Document arguments) {
```
adwsingh, line 94: This can be optimized by finding indexes where the replacement need to be done but we can take that up later on.

yasmewad: I see. Yeah I will add a TODO and create an issue for the same.

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpServer.java
yasmewad: Thanks for sharing that example it makes sense that we can use the `schema`. I will try that approach.

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/PromptLoader.java
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * Converts a Smithy structure shape to a list of PromptArgument objects.
+     *
+     * @param model The Smithy model containing the shape
+     * @param argumentShapeId The ShapeId of the structure to convert
+     * @return List of PromptArgument objects representing the structure members
+     */
+    public static List<PromptArgument> convertArgumentShapeToPromptArgument(Model model, ShapeId argumentShapeId) {
+        StructureShape argument = model.expectShape(argumentShapeId, StructureShape.class);
```
yasmewad, line 84: Is there a way to get this behavior using the `Service` class? I looked at the `TypeRegistry` and `Schema` but could not find a pattern on how to use it.

adwsingh: You will not have access to the StructureShape, but in this method all you are doing is accessing members and traits. Both can be done from the Schema class, using `Schema#getMembers` and `Schema#getTrait`.

yasmewad: I tested the suggested approach with @rhernandez35 and found a few issues:

Technical limitations discovered:
1. The McpServer builder uses List<Service> rather than List<ProxyService>, which could cause information loss
2. Neither Schema nor TypeRegistry contained the shapeId information needed for the shape  lookup like how it is easily available here.
3. The service Schema only contains the Service shape itself, and its members - no operations or resources, so this approach won't work for traits on those shapes. Even if we get it working for operations, there is no resource shape present in the Service schema or typeRegistry. The only other way is to use the SchemaConvertor to then access the shapeMap and get the shapes.  
4. TypeRegistry only has Error shapes (per ProxyService code), so we lack access to other shape information

Testing approach:
I updated ProxyMcpExample with traits in the model and added the suggested runtimeTraits method, then debugged through the Schema#getTrait loading process to confirm these 
limitations.

Alternative solutions we considered:
1. Parse the model directly in ProxyService.
2. Create prompts as operations with arguments ( I thought more on this but this exposes too much information and doesn't feel like the right translation in ProxyService)
3. Long-term I think we should move prompt parsing to a BuildPlugin so Smithy Build provides MCP-compliant objects that McpServer can consume at runtime. The McpServer can then still share the same logic of parsing with Build Plugin if we want. 

Given the complexity of reading StructureShape from the arguments member in the Prompts trait, I think passing the Model is the simplest viable approach for now.

Open to other suggestions if there's a simpler solution I'm missing.

adwsingh: > It only knows about the main service shape and its direct pieces. It doesn't include operations or resources (where our @prompts traits actually live) a valid use-case. So we can't even find the prompts, let alone process them

This isn't accurate. We generate Schema for both operations as well as resources. 

> The main blocker is ShapeId references. Prompts have an arguments field that points to a structure (like SearchArguments). To process this, we need to find that structure in the model and read its fields Schema and TypeRegistry just don't have this information

Should be fixed by #827

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/PromptLoader.java
```diff
@@ -0,0 +1,114 @@
...
+    public static Map<String, PromptInfo> loadPrompts(List<Service> services) {
+        Map<String, PromptInfo> promptInfos = new LinkedHashMap<>();
+
+        for (var service : services) {
+            Map<String, PromptTemplateDefinition> promptDefinitions = new HashMap<>();
+            var servicePromptTrait = service.schema().getTrait(PROMPTS_TRAIT_KEY);
+            if (servicePromptTrait != null) {
+                promptDefinitions.putAll(servicePromptTrait.getValues());
+            }
+            service.getAllOperations().forEach(operation -> {
```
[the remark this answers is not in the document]
adwsingh: Resources are converted to Operations, getAllOperations should represent resource operations too.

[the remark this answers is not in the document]
adwsingh: I don't think that will work, we do not copy traits on the resource to the individual operations, I think we might need some notion of resources in the Schema to access traits on resources. Let's tackle that in a different PR.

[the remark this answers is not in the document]
yasmewad: Yes. I can remove that capability in the trait definition for now. It will anyway be an additive change.

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpServer.java
```diff
@@ -113,12 +120,29 @@ private void handleRequest(JsonRpcRequest req) {
...
                                         .tools(Tools.builder().listChanged(true).build())
+                                        .prompts(Prompts.builder().listChanged(true).build())
                                         .build())
                                 .serverInfo(ServerInfo.builder()
                                         .name(name)
                                         .version("1.0.0")
                                         .build())
                                 .build());
+                case "prompts/list" -> writeResponse(req.getId(),
+                        ListPromptsResult.builder().prompts(prompts.values().stream().toList()).build());
```
rhernandez35, line 131: We need to prevent the prompt templates from being serialized in this response. We only want the name, title, description, and arguments.

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpServer.java
yasmewad: That's a valid point. Making it a default will save confusion at least for customers.

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/PromptLoader.java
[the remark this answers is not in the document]
yasmewad: This has been fixed in latest commits. Thanks @adwsingh

rhernandez35 (approved):

mcp/mcp-server/src/test/java/software/amazon/smithy/java/mcp/server/PromptProcessorTest.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Test
+    public void testApplyTemplateArgumentsWithMissingArgument() {
+        String template = "Hello {{name}}!";
+        Document arguments = Document.of(Map.of("other", Document.of("value")));
+
+        String result = processor.applyTemplateArguments(template, arguments);
+
+        assertEquals("Hello !", result);
```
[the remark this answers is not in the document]
yasmewad: Hmm, let me think about it. If there is `{{ name }}` but the arguments don't match, we can think of it as incorrect arguments passed by the modeler of the prompt not the user of the prompt. This is a trait validation thing from what I can think of.

We can definitely throw an error though, or return an error if there are arguments but no particular placeholder matches in the template.

Maybe a small update in separate PR.

PR #837, opened 2025-08-04
Add 'smithy.ai#prompts' as a default runtime trait'

rhernandez35:
It would be nice to bake this into the trait SPI directly instead of maintaining the static list.


codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/CodeGenerationContext.java
```diff
@@ -96,7 +96,9 @@ public class CodeGenerationContext
             IdempotencyTokenTrait.ID,
             RetryableTrait.ID,
             RequestCompressionTrait.ID,
-            StreamingTrait.ID);
+            StreamingTrait.ID,
+            // AI traits (experimental)
+            ShapeId.from("smithy.ai#prompts"));
```
adwsingh, line 101: So here it is. I was searching for this and couldn't find it.

adwsingh (approved):

PR #845, opened 2025-08-13
Make ImplicitErrorIndex immutable and deterministic

adwsingh:
*Issue #, if available:*

*Description of changes:*

The non-deterministic ordering of Shapes returned by getImplicitErrorsForService causes the TypeRegistry to be generated differently across runs. Since code generation should always be deterministic, producing identical output for the same model, this behavior is problematic.


rhernandez35:
sneaky!

rhernandez35 (approved):

PR #849, opened 2025-08-19
Event stream implementation for RPCv2

sugmanue:
*Description of changes:*

Initial wire of event streams for RPC protocols, and H2 support. This change only includes RPCv2, and a follow up change will add support for the other RPC protocols.

This change **does not** include any of:

1. Support for `@eventPayload` trait
2. Support for `@eventHeader` trait
3. Event signing

These will be added in follow up changes.

---

### DIFF
--- a/aws/aws-event-streams/build.gradle.kts
+++ b/aws/aws-event-streams/build.gradle.kts
@@ -9,5 +9,6 @@ extra["moduleName"] = "software.amazon.smithy.java.aws.events"
 
 dependencies {
     api(project(":core"))
+    api(project(":http:http-api"))
     api("software.amazon.eventstream:eventstream:1.0.1")
 }
--- a/aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/AwsEventDecoderFactory.java
+++ b/aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/AwsEventDecoderFactory.java
@@ -5,6 +5,7 @@
 
 package software.amazon.smithy.java.aws.events;
 
+import java.util.Objects;
 import java.util.function.Supplier;
 import software.amazon.smithy.java.core.schema.InputEventStreamingApiOperation;
 import software.amazon.smithy.java.core.schema.OutputEventStreamingApiOperation;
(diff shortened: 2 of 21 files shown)


examples/event-streaming-client/src/it/java/software/amazon/smithy/java/example/eventstreaming/EventStreamTest.java
```diff
@@ -85,16 +88,26 @@ public void onNext(FizzBuzzStream item) {
             }
 
             @Override
-            public void onError(Throwable throwable) {}
+            public void onError(Throwable throwable) {
+                System.out.println("Error: " + throwable.getMessage());
```
mtdowling, line 92: whoops

core/src/main/java/software/amazon/smithy/java/core/serde/event/EventDecoder.java
```diff
@@ -5,10 +5,20 @@
...
 
     SerializableStruct decode(F frame);
 
+    /**
+     * Called once after building the publisher to allow the decoder to do any one-time setup prior to start processing
+     * events.
+     *
+     * @param publisher The events publisher.
+     */
+    default void onPrepare(Flow.Publisher<SerializableStruct> publisher) {
```
mtdowling, line 21: Are decoders stateful?

aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/RpcEventStreamsRequest.java
```diff
@@ -0,0 +1,42 @@
...
+import software.amazon.smithy.java.core.schema.Schema;
+import software.amazon.smithy.java.core.schema.SerializableStruct;
+import software.amazon.smithy.java.core.schema.TraitKey;
+import software.amazon.smithy.java.core.serde.event.EventEncoderFactory;
+import software.amazon.smithy.java.core.serde.event.EventStreamFrameEncodingProcessor;
+
+/**
+ * A util class to serialize event streams requests for RPC protocols.
+ */
+public final class RpcEventStreamsRequest {
```
mtdowling, line 19: The name threw me off. Maybe RpcEventStreamsUtils would be more clear that this isn't actually a request, but request utils.

aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/RpcEventStreamResponse.java
```diff
@@ -0,0 +1,63 @@
...
+import software.amazon.smithy.java.core.schema.SerializableStruct;
+import software.amazon.smithy.java.core.serde.event.EventDecoderFactory;
+import software.amazon.smithy.java.core.serde.event.EventStreamFrameDecodingProcessor;
+import software.amazon.smithy.java.http.api.HttpResponse;
+import software.amazon.smithy.java.io.datastream.DataStream;
+
+/**
+ * A util class to deserialize event streams responses for RPC protocols.
+ */
+public final class RpcEventStreamResponse {
```
mtdowling, line 19: same thing here. Maybe even combine this utils class with the request utils class.

aws/aws-event-streams/build.gradle.kts
```diff
@@ -9,5 +9,6 @@ extra["moduleName"] = "software.amazon.smithy.java.aws.events"
 
 dependencies {
     api(project(":core"))
+    api(project(":http:http-api"))
```
mtdowling, line 12: Is this needed? I glanced but didn't see HTTP things actually getting used

aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/AwsEventDecoderFactory.java
```diff
@@ -17,43 +18,79 @@
...
+     * @param <OE>        The output event type
+     * @return A new event decoder factory
+     */
+    public static <OE extends SerializableStruct> AwsEventDecoderFactory<OE, ?> forOutputStream(
             OutputEventStreamingApiOperation<?, ?, OE> operation,
             Codec codec,
             FrameTransformer<AwsEventFrame> transformer
     ) {
         return new AwsEventDecoderFactory<>(
+                "initial-response",
```
mtdowling, line 92: Make this an enum?

mtdowling (changes requested):

core/src/main/java/software/amazon/smithy/java/core/serde/event/EventDecoder.java
sugmanue: Yes, decoders are stateful, but that's unrelated to this change: each keeps a reference to the schema of the event. 

This hook is here to support the initial response for RPC protocols which poses a bit of a chicken and egg situation:

* The initial response event get deserialized into the response object that the client gets, 
* The response also includes the member to read `Flow.Publisher` which is used by the caller to read the response events from the server
* The response is constructed out of reading of the publisher, and, the response provides access to the publisher
* This hook is then used to pass down the event processor back to the decoder such that it can put together the response with it
* Then to create the response we subscribe to the event publisher, and complete the response future with the initial request event.

aws/aws-event-streams/build.gradle.kts
sugmanue: > Is this needed? I glanced but didn't see HTTP things actually getting used

The RPC response util class has a method that takes a `HttpResponse` instance to create the API response. I will remove this dependency and the caller will need to create and pass down the data-stream.

aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/AwsEventDecoderFactory.java
sugmanue: Sure

aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/RpcEventStreamResponse.java
sugmanue: Sure.

aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/RpcEventStreamsRequest.java
sugmanue: Sure.

PR #852, opened 2025-08-21
Improve equals for floating point members

sugmanue:
For primitive floating-point members, the existing logic would render the equals method using `==`. This change uses instead `Float.compare` as it gives better results. E.g., for the added test model now the equals method looks like:

```java
    @Override
    public boolean equals(Object other) {
        if (other == this) {
            return true;
        }
        if (other == null || getClass() != other.getClass()) {
            return false;
        }
        StructureOne that = (StructureOne) other;
        return this.byteMember == that.byteMember
                && this.shortMember == that.shortMember
                && this.intMember == that.intMember
                && this.longMember == that.longMember
                && Float.compare(this.floatMember, that.floatMember) == 0
                && Double.compare(this.doubleMember, that.doubleMember) == 0
                && this.booleanMember == that.booleanMember;
    }
```

Also add a test runner to assert against expected generated java files. If the approach seems fine I can add more tests to the types codegen plugin and the client and server.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/generators/StructureGenerator.java
+++ b/codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/generators/StructureGenerator.java
@@ -60,6 +60,7 @@ import software.amazon.smithy.model.shapes.MapShape;
 import software.amazon.smithy.model.shapes.MemberShape;
 import software.amazon.smithy.model.shapes.ServiceShape;
 import software.amazon.smithy.model.shapes.Shape;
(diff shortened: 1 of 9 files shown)


sugmanue:
I guess the question is what's the `equals` method for. If we were working on a number crunching library I would agree with you, however, I feel like the users of libraries like this one will benefit a lot more from a behavior that mirrors intuition. Strict  adherence to IEEE-754 starts as surprising and ends up being just an annoyance.

sugmanue:
The team decided to keep the current behavior instead.

PR #855, opened 2025-08-23
Add a plugin test runner and a test for types-codegen simple types

sugmanue:
*Issue #, if available:*

*Description of changes:*

This change adds a utility class `PluginTestRunner` that can be used by all codegen plugins to validate the code being generated. 

Also added a single test for the `TypeCodegenPlugin`. 

**The below issue was fixed in #856**

This test shows a currently existing bug in the builders for nullable types, the setters are codegen to take primitive types which will make `.toBuilder()` to fail if a nullable type is null and will make it impossible to update a member from a non-null value to a null one. E.g., this will fail:

```
   PrimitivesNullable.builder()
        .longMember(10L)
        .build()
        .toBuilder()
        .longMember(null)
        .build();
```

I will fix that issue after merging this change.

### DIFF
--- a/codegen/plugins/types-codegen/src/test/java/software/amazon/smithy/java/codegen/types/TypesCodegenPluginTest.java
+++ b/codegen/plugins/types-codegen/src/test/java/software/amazon/smithy/java/codegen/types/TypesCodegenPluginTest.java
@@ -40,8 +40,7 @@ public class TypesCodegenPluginTest {
     }
 
     // Uncomment this test to render the java files when we there are changes to the codegen logic.
-    @ParameterizedTest(name = "[{index}] => {0}")
-    @MethodSource("testCases")
+    // @ParameterizedTest(name = "[{index}] => {0}") @MethodSource("testCases")
     public void renderExpected(TestCase test) throws IOException {
         test.builder().build();
         var got = test.manifests().stream().flatMap(x -> x.getFiles().stream()).collect(Collectors.toSet());


codegen/test-utils/src/main/java/software/amazon/smithy/java/codegen/test/PluginTestRunner.java
```diff
@@ -0,0 +1,258 @@
...
+            }
+        }
+        return Optional.empty();
+    }
+
+    public static void assertContentEquals(String left, String right) {
+        try {
+            assertEquals(normalizeSpace(left), normalizeSpace(right));
+        } catch (Throwable e) {
+            assertEquals(left, right);
```
adwsingh, line 62: IMO we shouldn’t apply any normalization. Codegen must produce identical output, byte-for-byte, every time.

codegen/test-utils/src/main/java/software/amazon/smithy/java/codegen/test/PluginTestRunner.java
```diff
@@ -0,0 +1,258 @@
...
+    public static class JavaFileVisitor extends SimpleFileVisitor<Path> {
+        private final List<Path> javaFiles;
+
+        public JavaFileVisitor(List<Path> javaFiles) {
+            this.javaFiles = javaFiles;
+        }
+
+        @Override
+        public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
+            if (Files.isRegularFile(file) && file.toString().endsWith(".java")) {
```
adwsingh, line 245: Is there a reason to only load `.java` files? Codegen also produces property and SPI config files (`META-INF/services`).

adwsingh (changes requested):

codegen/test-utils/src/main/java/software/amazon/smithy/java/codegen/test/PluginTestRunner.java
sugmanue: I thought I removed this bit as discussed, my bad, let me fix it.

codegen/test-utils/src/main/java/software/amazon/smithy/java/codegen/test/PluginTestRunner.java
sugmanue: No particular reason, I can remove this bit.

codegen/plugins/types-codegen/src/test/java/software/amazon/smithy/java/codegen/types/TypesCodegenPluginTest.java
```diff
@@ -0,0 +1,64 @@
...
+            var found = findExpected(expected, got);
+            assertNotNull(found, "Expected to find " + expected + " in the manifest");
+            var contents = findGotContent(found, test);
+            assertTrue(contents.isPresent());
+            assertEquals(test.expectedToContents().get(expected), contents.get());
+        }
+    }
+
+    // Uncomment this test to render the java files when we there are changes to the codegen logic.
+    @ParameterizedTest(name = "[{index}] => {0}") @MethodSource("testCases")
```
adwsingh: I think you meant to comment this.

adwsingh (approved):

codegen/plugins/types-codegen/src/test/java/software/amazon/smithy/java/codegen/types/TypesCodegenPluginTest.java
sugmanue: Yes, you're right. Let me fix that.

PR #865, opened 2025-09-03
Delay publishing the initial event until fully wired

sugmanue:
There's a race condition that can happen when we add the initial request using onNext, that call will call flush while the downstream subscriber is not yet set. To avoid this we just enqueue the initial request, and wait till the upstream requests more input.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/RpcEventStreamsUtil.java
+++ b/aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/RpcEventStreamsUtil.java
@@ -31,7 +31,7 @@ public final class RpcEventStreamsUtil {
         Flow.Publisher<SerializableStruct> eventStream = input.getMemberValue(streamingMember(input.schema()));
         var publisher = EventStreamFrameEncodingProcessor.create(eventStream, eventStreamEncodingFactory);
         // Queue the input as the initial-request.
-        publisher.onNext(input);
+        publisher.enqueueItem(input);
         return publisher;
     }
 
@@ -62,7 +62,7 @@ public final class RpcEventStreamsUtil {
 
             @Override
             public void onComplete() {
-                result.completeExceptionally(new RuntimeException("Unexpected vent stream completion"));
+                result.completeExceptionally(new RuntimeException("Unexpected event stream completion"));
             }
         });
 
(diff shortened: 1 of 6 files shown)


mtdowling (approved):

core/src/main/java/software/amazon/smithy/java/core/serde/BufferingFlatMapProcessor.java
```diff
@@ -64,13 +64,23 @@ public final void subscribe(Flow.Subscriber<? super O> subscriber) {
...
+        flush();
+    }
+
+    /**
+     * Adds the item to the queue for later delivery. The queue will get flushed upon calls to {@link #request} or
+     * {@link #onNext(Object)}.
+     *
+     * @param item The item to add to the queue
+     */
+    public final void enqueueItem(I item) {
```
rhernandez35, line 77: We can not add a new API that must be used for proper flow outside of the rx spec

aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/RpcEventStreamsUtil.java
```diff
@@ -31,7 +31,7 @@ public static Flow.Publisher<ByteBuffer> bodyForEventStreaming(
         Flow.Publisher<SerializableStruct> eventStream = input.getMemberValue(streamingMember(input.schema()));
         var publisher = EventStreamFrameEncodingProcessor.create(eventStream, eventStreamEncodingFactory);
         // Queue the input as the initial-request.
-        publisher.onNext(input);
+        publisher.enqueueItem(input);
```
rhernandez35, line 34: Presumably, you are doing it this way because we may not have been subscribed to yet. If so, our reactive code is broken elsewhere (maybe here?) and must be brought back into compliance with the reactive streaming specification.

rhernandez35 (changes requested):

core/src/main/java/software/amazon/smithy/java/core/serde/BufferingFlatMapProcessor.java
```diff
@@ -64,13 +64,17 @@ public final void subscribe(Flow.Subscriber<? super O> subscriber) {
 
     @Override
     public final void onNext(I item) {
+        enqueueItem(item);
+        flush();
```
rhernandez35, line 68: This should not be called if `onError` was invoked in `enqueueItem`

rhernandez35 (approved):

PR #878, opened 2025-09-24
Add output schema to MCP tools/list

rhernandez35:
Part of me thinks this should only be enabled with a CLI option given the impact it'll have on the context window.


mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpServer.java
```diff
@@ -339,6 +339,8 @@ private Map<String, Tool> createTools(Map<String, Service> services) {
                                 operationName,
                                 schema))
                         .inputSchema(createJsonObjectSchema(operation.getApiOperation().inputSchema(), new HashSet<>()))
+                        .outputSchema(
```
adwsingh, line 342: I think we should do this based on the MCP version. This was only added in version `2024-11-05`

PR #902, opened 2025-10-16
Improve name conflicts resolving

joewyz:
*Description of changes:*
This PR is aimed to fix two problems:
1. Current implementation added `type` into the `reserved-words.txt` and this approach is too broad. 
We want to only resolve the conflict when the name `type` is used, for example, `Union` or `Enum`.

2. The original approach to resolve implicit name conflict tried to add all symbols generated under the same package into each file's `symbolTable` which caused some unnecessary name conflicts. 
We only want to resolve the conflicts when those auto-imported `java.lang.*` classes like `illegalArgumentException` are used and one class with the same name defined under the same namespace.

For example, the `ShapeSerializer` is used with a fully qualified name even though it is imported. This happened because one shape also called `ShapeSerializer` was generated under the same namespace and added into this file's symbol table. 
```java
import software.amazon.smithy.java.core.serde.ShapeSerializer;

...

@Override
public void serializeMembers(software.amazon.smithy.java.core.serde.ShapeSerializer serializer) {

}
```

### DIFF
--- a/codegen/plugins/types-codegen/src/test/resources/software/amazon/smithy/java/codegen/types/test-cases/naming-conflict/model/main.smithy
+++ b/codegen/plugins/types-codegen/src/test/resources/software/amazon/smithy/java/codegen/types/test-cases/naming-conflict/model/main.smithy
@@ -36,8 +36,6 @@ structure Builder {}
 
 structure Type {}
 
-// All of the members of this structure would override
-// Object.class method's unless escaped.
 structure Object {
     class: String
     getClass: String
@@ -50,13 +48,10 @@ structure Object {
     finalize: String
 }
 
-@private
 union UnionWithTypeMember {
     type: Type
 }
 
-@private
 structure Map {}
 
-@private
 structure List {}


codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/generators/UnionGenerator.java
```diff
@@ -42,6 +43,8 @@ public void accept(GenerateUnionDirective<CodeGenerationContext, JavaCodegenSett
         }
         var shape = directive.shape();
         directive.context().writerDelegator().useShapeWriter(shape, writer -> {
+            Symbol innerTypeEnumSymbol = CodegenUtils.getInnerTypeEnumSymbol(directive.symbol());
```
sugmanue: Prefer `var` over explicit type of the variable.

codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/writer/JavaWriter.java
```diff
@@ -87,28 +89,54 @@ public void newLine() {
...
+    }
+
+    /**
+     * Records the local name defined within the current file, e.g., the name of an inner class.
+     * The writer will use this information to decide if it should use a fully qualified
+     * name when referencing types with the same name in its scope.
+     *
+     * @param symbol the symbol reserved in this file.
+     */
+    public void addLocalDefinedSymbol(Symbol symbol) {
```
sugmanue:
```suggestion
    public void addLocallyDefinedSymbol(Symbol symbol) {
```

codegen/plugins/types-codegen/src/test/resources/software/amazon/smithy/java/codegen/types/test-cases/naming-conflict/model/main.smithy
```diff
@@ -0,0 +1,52 @@
...
+}
+
+@private
+structure Builder {}
+
+@private
+structure Type {}
+
+// All of the members of this structure would override
+// Object.class method's unless escaped.
```
sugmanue: After the change to use java getters, this statement is not longer true. Let's remove it.

codegen/plugins/types-codegen/src/test/resources/software/amazon/smithy/java/codegen/types/test-cases/naming-conflict/model/main.smithy
```diff
@@ -0,0 +1,52 @@
...
+@private
+union UnionWithTypeMember {
+    type: Type
+}
+
+@private
+structure Map {}
+
+@private
+structure List {}
```
sugmanue: Let’s just use these structures, `Map` and `List` in a context where we also have `java.util.{List,Map}` imported. We want to validate that in that case there will be no name conflicts.

Issue #913, opened 2025-11-04
File Descriptor leak in JavaHttpClientTransport


cdj-offside:
Noticed in one of my services that memory was growing constantly and would eventually reach OOM. Eventually found that it was a socket FD leak, not 100% sure why (I don't know enough about Java's client implementation), but replacing it with an Apache client transport resolved the issue.

This is version 0.0.3, not sure if it's been fixed since then, but just wanted to call it out.

mtdowling:
Can you check to see if #916 addresses this, and if not, share more about how and what you're calling? Is a payload being sent from the client, is the service returning a payload? Are you consuming it entirely?

PR #916, opened 2025-11-06
Orient DataStream around InputStream

mtdowling:
DataStream now is written around the idea of using an underyling InputStream more often than a Publisher, so default implementations were updated (now default provided for asPublisher, and asInputStream no longer has a default).

Also closes InputStreams in a couple places that we could have leaked.

Addresses #913

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/io/src/main/java/software/amazon/smithy/java/io/datastream/InputStreamDataStream.java
+++ b/io/src/main/java/software/amazon/smithy/java/io/datastream/InputStreamDataStream.java
@@ -5,20 +5,14 @@
 
 package software.amazon.smithy.java.io.datastream;
 
-import java.io.IOException;
 import java.io.InputStream;
-import java.io.UncheckedIOException;
-import java.net.http.HttpRequest;
-import java.nio.ByteBuffer;
-import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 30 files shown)


rhernandez35 (approved):

io/src/main/java/software/amazon/smithy/java/io/datastream/InputStreamDataStream.java
```diff
@@ -27,17 +21,12 @@ final class InputStreamDataStream implements DataStream {
     }
 
     @Override
-    public ByteBuffer waitForByteBuffer() {
-        try {
-            return ByteBuffer.wrap(inputStream.readAllBytes());
-        } catch (IOException e) {
-            throw new UncheckedIOException(e);
+    public InputStream asInputStream() {
+        if (consumed) {
```
adwsingh: FYI, this isn't thread safe, I am not entirely sure if that's acceptable here.

mtdowling: I figured that InputStream and reading data is entirely not thread safe, so it was fine

PR #922, opened 2025-11-11
Add error response parsing for restXml

joewyz:
Current implementation for restXml could not parse the [error response](https://smithy.io/2.0/aws/protocols/aws-restxml-protocol.html#error-response-serialization) because the HttpErrorDeserializer is converting the response to document which is not supported in restXml. This PR adds a fix for this to pass protocol tests "InvalidGreetingError" and "ComplexError".

### DIFF
--- a/codecs/xml-codec/src/main/java/software/amazon/smithy/java/xml/XmlDeserializer.java
+++ b/codecs/xml-codec/src/main/java/software/amazon/smithy/java/xml/XmlDeserializer.java
@@ -132,7 +132,7 @@ final class XmlDeserializer implements ShapeDeserializer {
             var element = reader.nextMemberElement();
             if (element == null || (!element.equals("ErrorResponse") && !element.equals("Error"))) {
                 throw new SerializationException(
-                        "Expected element <ErrorResponse> or <Error> for restXml error response");
+                        "Expected element <ErrorResponse> or <Error> for XML error response");
             }
             if (element.equals("ErrorResponse")) {
                 element = reader.nextMemberElement();
--- a/codecs/xml-codec/src/main/java/software/amazon/smithy/java/xml/XmlUtil.java
+++ b/codecs/xml-codec/src/main/java/software/amazon/smithy/java/xml/XmlUtil.java
@@ -12,7 +12,7 @@ import software.amazon.smithy.java.core.serde.ShapeDeserializer;
  */
 public final class XmlUtil {
     /**
-     * Retrieve the Code element value from the error response.
+     * Parse the Code element value from error response for both AWS restXml and AWS Query protocols.
      *
      * @param deserializer the deserializer for the error response
      * @return String value of the Code element if found


client/client-http/src/main/java/software/amazon/smithy/java/client/http/HttpErrorDeserializer.java
```diff
@@ -213,6 +214,41 @@ public CallException createError(
...
+    /**
+     * Create the error from response for restXml Protocol specifically.
+     *
+     * @param context      Context for the request.
+     * @param operation    Operation to create request for.
+     * @param typeRegistry TypeRegistry that can be used to create shapes.
+     * @param response     Response to deserialize.
+     * @return the deserialized error response.
+     */
+    public CallException createErrorForXml(
```
mtdowling: HttpErrorDeserializer is not protocol specific, so it shouldn't know anything specific to XML here

codecs/xml-codec/src/main/java/software/amazon/smithy/java/xml/XmlCodec.java
```diff
@@ -65,6 +66,27 @@ public ShapeDeserializer createDeserializer(ByteBuffer source) {
         }
     }
 
+    private XmlDeserializer createXmlDeserializer(ByteBuffer source) {
```
mtdowling: I think this is mostly duplicative of the above method

client/client-http/src/main/java/software/amazon/smithy/java/client/http/HttpErrorDeserializer.java
```diff
@@ -362,5 +409,18 @@ public Builder unknownErrorFactory(UnknownErrorFactory unknownErrorFactory) {
             this.unknownErrorFactory = Objects.requireNonNull(unknownErrorFactory, "unknownErrorFactory is null");
             return this;
         }
+
+        /**
+         * The parser to parse the shapeId from the payload.
+         *
+         * <p>By default, payload is parsed into a JSON document
```
mtdowling: Hm that's not right. It's parsed into a document I think, and documents are not == JSON. Please also expand on how it identifies an error by using the discriminator of a document.

codecs/xml-codec/src/main/java/software/amazon/smithy/java/xml/XmlDeserializer.java
```diff
@@ -80,6 +80,9 @@ private void enter(Schema schema) {
                 expected = trait.getValue();
             } else if (schema.isMember()) {
                 expected = schema.memberTarget().id().getName();
+            } else if (name != null && (name.equals("ErrorResponse") || name.equals("Error"))) {
+                skipToCodeElement(name);
```
mtdowling: zoinks. can you give an in-code example of what this is doing and show the XML that leads to this? Also we return early here -- why?

joewyz: An ErrorResponse will look like this:
```xml
<ErrorResponse>
    <Error>
        <Type>Sender</Type>
        <Code>InvalidGreeting</Code>
        <Message>Hi</Message>
        <AnotherSetting>setting</AnotherSetting>
    </Error>
    <RequestId>foo-id</RequestId>
</ErrorResponse>
```
I think when we try to deserialize the error, `readStruct` will be triggered and it will do `enter()`first. This `enter` function will throw exception if we do not return early since `ErrorResponse` is not equal to the error shape name. Then we should skip all the way to the code element and deserialize the rest elements.

codecs/xml-codec/src/main/java/software/amazon/smithy/java/xml/XmlDeserializer.java
```diff
@@ -66,6 +66,16 @@ public void close() {
 
     // The first deserialization of XML expects a containing XML element for the shape.
     // The inner deserializer deserializes members and doesn't have this expectation.
+    // If the name is ErrorResponse or Error, it means we are deserializing an ErrorResponse like:
```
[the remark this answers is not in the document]
joewyz: I think from [our doc here](https://smithy.io/2.0/aws/protocols/aws-query-protocol.html#operation-error-serialization), this fix should also work for aws query protocol.

mtdowling (approved):

PR #931, opened 2025-11-14
Optimize Context implementation with chunked array storage

mtdowling:
Replace map-based and separate array-based Context storage with a single chunked array implementation that provides significant performance improvements over the map-based implementation across all operations while maintaining memory efficiency through lazy allocation. Because we're already around 50 context keys just in the main repo, the array-based storage is essentially a dead-end and we need to be smarter about allocations for contexts.

Chunked storage vs map storage:
- 3.1x faster context copying
- 2.5x faster random access
- 1.5x faster on all put/get operations
- Eliminates potentially megamorphic call sites by using single storage implementation
- Scales efficiently from 1 to 1000+ keys with 32-element chunks

Benchmarks were added to compare an approximate map-based impl. vs the chunked impl. Results as follows:

```
Benchmark                                   Mode  Cnt    Score   Error  Units
ContextBenchmark.copyContext                avgt    5   40.730 ± 2.675  ns/op
ContextBenchmark.createEmptyContext         avgt    5    3.658 ± 0.045  ns/op
ContextBenchmark.getFirstKey                avgt    5    1.242 ± 0.072  ns/op
ContextBenchmark.getLastKey                 avgt    5    1.232 ± 0.028  ns/op
ContextBenchmark.getMissingKey              avgt    5    1.286 ± 0.107  ns/op
ContextBenchmark.getRandomKey               avgt    5    2.483 ± 0.076  ns/op
ContextBenchmark.putMultipleKeys            avgt    5   21.193 ± 0.577  ns/op
ContextBenchmark.putRandomKey               avgt    5    9.911 ± 0.268  ns/op
MapBench.copyContext                        avgt    5  127.043 ± 0.684  ns/op
MapBench.createEmptyContext                 avgt    5    3.205 ± 0.120  ns/op
MapBench.getFirstKey                        avgt    5    1.912 ± 0.206  ns/op
MapBench.getLastKey                         avgt    5    3.246 ± 0.131  ns/op
MapBench.getMissingKey                      avgt    5    1.510 ± 0.114  ns/op
MapBench.getRandomKey                       avgt    5    6.334 ± 0.199  ns/op
MapBench.putMultipleKeys                    avgt    5   32.575 ± 0.334  ns/op
MapBench.putRandomKey                       avgt    5   14.882 ± 0.100  ns/op
```

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/context/src/main/java/software/amazon/smithy/java/context/Context.java
+++ b/context/src/main/java/software/amazon/smithy/java/context/Context.java
@@ -5,9 +5,9 @@
 
 package software.amazon.smithy.java.context;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.Function;
 
 /**
@@ -26,9 +26,13 @@ public sealed interface Context permits ChunkedArrayStorageContext, Unmodifiable
      */
     final class Key<T> {
 
-        // Global registry of all keys for copyTo operations
+        // Global registry of all keys for copyTo operations (CopyOnWriteArrayList for thread safety).
         @SuppressWarnings("rawtypes")
-        static final List<Key> KEYS = new ArrayList<>();
+        static final List<Key> KEYS = new CopyOnWriteArrayList<>();
+
+        // We still need a lock to atomically insert an ID into KEYS and get the ID.
+        // Spotbugs warns against Synchronizing on KEYS, so using a dedicated lock.
+        private static final Object KEYS_LOCK = new Object();
 
         private final String name;
         final int id;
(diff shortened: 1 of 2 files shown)


context/src/main/java/software/amazon/smithy/java/context/Context.java
```diff
@@ -45,8 +43,13 @@ final class Key<T> {
          */
         private Key(String name, Function<T, T> copyFunction) {
             this.name = Objects.requireNonNull(name);
-            this.id = COUNTER.getAndIncrement();
             this.copyFunction = Objects.requireNonNull(copyFunction);
+
+            // Atomically assign our slot in the KEYS array.
+            synchronized (KEYS_LOCK) {
```
rhernandez35: The`synchronized` isn't necessary any more

PR #932, opened 2025-11-14
Add document discriminator sanitizer for JSON protocols

joewyz:
This PR includes multiple changes:

1. added protocol tests for AWS JSON 1.1 
2. allowed awsJson protocols to use TimestampResolver
3. added error type sanitizer for JSON protocols given we are using `__type` as the error's shapeId and different JSON protocols have different requirement for `__type`'s format. 
With this fix, those protocol tests with `__type` requirement can pass.

### DIFF
--- a/codecs/json-codec/src/main/java/software/amazon/smithy/java/json/JsonSettings.java
+++ b/codecs/json-codec/src/main/java/software/amazon/smithy/java/json/JsonSettings.java
@@ -7,7 +7,6 @@ package software.amazon.smithy.java.json;
 
 import java.util.Objects;
 import java.util.ServiceLoader;
-import java.util.function.Function;
 import software.amazon.smithy.java.core.serde.TimestampFormatter;
 
 /**
@@ -46,7 +45,6 @@ public final class JsonSettings {
     private final JsonSerdeProvider provider;
     private final boolean serializeTypeInDocuments;
     private final boolean prettyPrint;
-    private final Function<String, String> errorTypeSanitizer;
 
     private JsonSettings(Builder builder) {
         this.timestampResolver = builder.useTimestampFormat
@@ -60,7 +58,6 @@ public final class JsonSettings {
         this.provider = builder.provider;
         this.serializeTypeInDocuments = builder.serializeTypeInDocuments;
         this.prettyPrint = builder.prettyPrint;
-        this.errorTypeSanitizer = builder.errorTypeSanitizer;
     }
 
     /**
@@ -119,16 +116,6 @@ public final class JsonSettings {
     }
 
     /**
-     * The error type sanitizer to use for {@code __type} or {@code code} when parsing the error discriminator.
-     * Default is null
-     *
-     * @return the sanitizer used or null
-     */
-    public Function<String, String> errorTypeSanitizer() {
-        return errorTypeSanitizer;
-    }
-
-    /**
      * Whether to format the JSON output with pretty printing (indentation and line breaks).
      *
      * @return true if pretty printing is enabled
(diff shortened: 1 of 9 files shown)


mtdowling (changes requested):

codecs/json-codec/src/main/java/software/amazon/smithy/java/json/ErrorTypeSanitizer.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * A util class contains the sanitizer functions to process {@code __type} field from the document
+ *
+ * <p>For example, given {@code __type = "aws.protocoltests.restjson#FooError:http://abc.com"},
+ * different protocols have different requirements for the {@code __type} field. This class provides
+ * sanitizers to remove trailing URIs or leading namespaces from {@code __type}.
+ */
+public class ErrorTypeSanitizer {
+    public static String REMOVE_URI(String text) {
```
sugmanue: can we use regular naming here, i.e., `removeUri` and `removeNamespaceAndUri`?

codecs/json-codec/src/main/java/software/amazon/smithy/java/json/ErrorTypeSanitizer.java
```diff
@@ -0,0 +1,37 @@
...
+        if (text == null) {
```
sugmanue: Can we assume that the caller is handling that instead and just `Objects.requireNotNull(…)` here?

aws/client/aws-client-awsjson/src/main/java/software/amazon/smithy/java/aws/client/awsjson/AwsJsonProtocol.java
```diff
@@ -39,10 +40,14 @@ abstract sealed class AwsJsonProtocol extends HttpClientProtocol permits AwsJson
...
      */
-    public AwsJsonProtocol(ShapeId trait, ShapeId service) {
+    public AwsJsonProtocol(ShapeId trait, ShapeId service, Function<String, String> errorTypeSanitizer) {
         super(trait);
         this.service = service;
-        this.codec = JsonCodec.builder().defaultNamespace(service.getNamespace()).build();
+        this.codec = JsonCodec.builder()
+                .defaultNamespace(service.getNamespace())
+                .errorTypeSanitizer(errorTypeSanitizer)
+                .useTimestampFormat(true)
```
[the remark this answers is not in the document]
joewyz: we have some tests like [this](https://github.com/smithy-lang/smithy/blob/fb91fea3d07fbe8b79769bdbd7461bea9473dcf8/smithy-aws-protocol-tests/model/awsJson1_1/kitchen-sink.smithy#L200) in 1.1's protocol tests

codecs/json-codec/src/main/java/software/amazon/smithy/java/json/ErrorTypeSanitizer.java
joewyz: The caller is not handling it, but would it be better to make the caller to handle null check?

codecs/json-codec/src/main/java/software/amazon/smithy/java/json/JsonSettings.java
```diff
@@ -115,6 +118,15 @@ public boolean serializeTypeInDocuments() {
         return serializeTypeInDocuments;
     }
 
+    /**
+     * The error type sanitizer to use for {@code __type}. Default is null
```
mtdowling: Hm, __type can be used for more than just error discrimination though. Maybe this should be more just a typeSanitizer.

mtdowling (changes requested):

PR #934, opened 2025-11-15
add McpMetricsObserver

sanskritipatole:
*Issue #, if available:*

*Description of changes:*

Add metric observer for client requests:

Captures :
• methodName - `tools/call` || `initialize`    
• protocolVersion - if present, null otherwise
• rootsListChanged - true if capabilities.roots.listChanged is true, false otherwise
• sampling - true if capabilities.y exists, false otherwise
• elicitation - true if capabilities.elicitation exists, false otherwise
• clientName - extracted if clientInfo.name exists, null otherwise
• clientTitle - extracted if clientInfo.title exists, null otherwise


mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpService.java
```diff
@@ -112,6 +115,45 @@ public JsonRpcResponse handleRequest(
             Consumer<JsonRpcResponse> asyncResponseCallback,
             ProtocolVersion protocolVersion
     ) {
+        if (metricsObserver != null) {
+            try {
+                var params = req.getParams();
+                var clientInfo = params.getMember("clientInfo");
+                var capabilities = params.getMember("capabilities");
+
```
adwsingh: This will always be null if its not an `initialize` call.

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpService.java
```diff
@@ -112,6 +115,52 @@ public JsonRpcResponse handleRequest(
             Consumer<JsonRpcResponse> asyncResponseCallback,
             ProtocolVersion protocolVersion
     ) {
+        if (metricsObserver != null) {
+            try {
+                if ("initialize".equals(req.getMethod())) {
```
adwsingh: You don't need to do any matching here. We already do it below just add these to the methods there.

adwsingh (approved):

PR #943, opened 2025-11-21
feat: Add support for native remote MCP servers

DivyaMaddipudi:
*Issue #, if available:*

*Description of changes:*

This commit adds support for native remote MCP servers with the following key changes:
- Proxies are now started immediately but initialized only after receiving an actual initialize request
- startProxies() starts proxy connections while initializeProxies() handles the MCP protocol initialization
- Uses proxiesInitialized flag to prevent duplicate initialization

### DIFF
--- a/mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpServerProxy.java
+++ b/mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpServerProxy.java
@@ -23,7 +23,7 @@ public abstract class McpServerProxy {
     private static final AtomicInteger ID_GENERATOR = new AtomicInteger(0);
 
     protected Consumer<JsonRpcResponse> notificationConsumer;
-    protected String protocolVersion;
+    protected volatile String protocolVersion;
 
     public List<ToolInfo> listTools() {
         JsonRpcRequest request = JsonRpcRequest.builder()


mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/HttpMcpProxy.java
```diff
@@ -0,0 +1,144 @@
...
+
+    private final HttpClient client;
+    private final URI endpoint;
+    private final String name;
+    private final Map<String, String> headers;
+
+    private HttpMcpProxy(Builder builder) {
+        this.client = HttpClient.newBuilder()
+                .connectTimeout(Duration.ofSeconds(30))
+                .build();
```
adwsingh: Use the smithy-java ClientTransport, not the Java HttpClient directly.

adwsingh (changes requested):

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpService.java
```diff
@@ -279,23 +289,38 @@ private JsonRpcResponse handleToolsCall(
...
 
     /**
-     * Initializes proxies with the given response writer and initialize request.
+     * Starts proxies without initializing them.
      */
-    public void initializeProxies(Consumer<JsonRpcResponse> responseWriter) {
+    public void startProxies() {
         for (McpServerProxy proxy : proxies.values()) {
-            proxy.start();
-            proxy.initialize(responseWriter, initializeRequest.get());
```
DivyaMaddipudi: initializeRequest.get() is giving me an NPE because we don't have any JSON-RPC request when the proxy starts. We need to either manually pass a JSON-RPC request to initialize the call or perform the initialization. For example: 

https://github.com/smithy-lang/smithy-java/pull/943/files#diff-98c1d22a8180d3700e143f1b567faab2c3d3bdfac31593242af08bf77ce1d616R171

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/HttpMcpProxy.java
```diff
@@ -0,0 +1,165 @@
...
+                if (response.statusCode() != 200) {
+                    throw new RuntimeException("HTTP error " + response.statusCode());
+                }
+
+                return JsonRpcResponse.builder()
+                        .deserialize(JSON_CODEC.createDeserializer(response.body().asByteBuffer()))
+                        .build();
+            } catch (Exception e) {
+                LOG.error("HTTP request failed", e);
+                throw new RuntimeException("HTTP request failed: " + e.getMessage(), e);
```
adwsingh: Don't log and throw, just throw.

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/HttpMcpProxy.java
```diff
@@ -0,0 +1,165 @@
...
+                    throw new IllegalArgumentException("JsonRpcRequest cannot be null");
+                }
+
+                byte[] body = JSON_CODEC.serializeToString(request).getBytes(StandardCharsets.UTF_8);
+                LOG.trace("Sending HTTP request to {}", endpoint);
+
+                HttpRequest.Builder requestBuilder = HttpRequest.builder()
+                        .uri(endpoint)
+                        .method("POST")
+                        .withAddedHeader("Content-Type", "application/json")
```
adwsingh: You need to send a Accept header, which sends both `application/json` and `text/event-stream`

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/HttpMcpProxy.java
```diff
@@ -0,0 +1,165 @@
...
+                    requestBuilder.withAddedHeader(header.getKey(), header.getValue());
+                }
+
+                Context context = Context.create();
+                context.put(HttpContext.HTTP_REQUEST_TIMEOUT, timeout);
+
+                HttpResponse response = transport.send(context, requestBuilder.build());
+                LOG.trace("Received HTTP response with status: {}", response.statusCode());
+
+                if (response.statusCode() != 200) {
```
adwsingh: This can also be a 202. You should check if its a 2xx status code.

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpService.java
```diff
@@ -279,23 +289,40 @@ private JsonRpcResponse handleToolsCall(
...
+     * Initializes proxies with the actual initialize request.
+     */
+    public void initializeProxies(Consumer<JsonRpcResponse> responseWriter) {
+        JsonRpcRequest initRequest = initializeRequest.get();
+        if (initRequest == null) {
+            LOG.warn("Cannot initialize proxies: no initialize request received yet");
+            return;
+        }
+
+        System.out.println("DEBUG: Initializing " + proxies.size() + " proxies");
```
DivyaMaddipudi: Remove these in the follow-up commit

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/HttpMcpProxy.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    private static String sanitizeName(String host) {
+        if (host == null) {
+            return "http-proxy-mcp";
+        }
+        return host.replaceAll("[^a-zA-Z0-9-]", "-");
+    }
+
+    public static class Builder {
```
adwsingh:
```suggestion
    public static final class Builder {
```

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/HttpMcpProxy.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    @Override
+    public CompletableFuture<JsonRpcResponse> rpc(JsonRpcRequest request) {
+        return CompletableFuture.supplyAsync(() -> {
```
adwsingh: Don't run this asynchronously, its gonna make debugging a pain. Just run synchronously and return a completed future.

adwsingh (changes requested):

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/HttpMcpProxy.java
```diff
@@ -0,0 +1,185 @@
...
+            String contentType = response.body().contentType();
+            byte[] bodyBytes = ByteBufferUtils.getBytes(response.body().asByteBuffer());
+
+            if (contentType != null && contentType.contains("application/json")) {
+                try {
+                    return JsonRpcResponse.builder()
+                            .deserialize(JSON_CODEC.createDeserializer(bodyBytes))
+                            .build();
+                } catch (Exception e) {
+                    LOG.warn("Failed to deserialize JSON error response", e);
```
adwsingh: If its an invalid JSON or a not a invalid RPC response, return a JsonRpc Response stating with the error that its an Invalid Json, there is no point sending the invalid JSON to the LLM.

The below piece of code where we print the first 200 bytes, we solely do this so that things like 502 Bad Gateway or similar errors from the load balancer can be surfaced. If the application is sending an invalid JSON response, we don't display it at all.

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/HttpMcpProxy.java
```diff
@@ -0,0 +1,185 @@
...
+                }
+            }
+
+            String errorBody = new String(bodyBytes, 0, Math.min(200, bodyBytes.length), StandardCharsets.UTF_8);
+            errorMessage = errorMessage + ": " + errorBody;
```
adwsingh: Explicitly show that this is truncated if we are picking the first 200 bytes only.

Something like

```
int length = Math.min(200, bodyBytes.length)

errorMessage = errorMessage + ": " + errorBody + (length != bodyBytes.length ? " (truncated)" + "";
```

adwsingh (changes requested):

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpServerProxy.java
```diff
@@ -23,6 +23,7 @@ public abstract class McpServerProxy {
     private static final AtomicInteger ID_GENERATOR = new AtomicInteger(0);
 
     protected Consumer<JsonRpcResponse> notificationConsumer;
+    protected String protocolVersion;
```
adwsingh: This needs to be volatile, other wise its not guaranteed to be visible to other threads in subsequent requests.

adwsingh (approved):

PR #945, opened 2025-11-24
Fix httpQuery related protocol tests

### LINKED ISSUE #937: @httpQuery failed to work when it is applied to list member in generated client
When `@httpQuery` trait is applied to a list member, the generated client failed to include it in the request URL. Here is a simple model can be used to reproduce the error:
```smithy
$version: "2"

namespace foo.bar

service TestService {
    operations: [
        Echo
    ]
}

@http(method: "GET", uri: "/")
operation Echo {
    input: EchoInput
    output: EchoOutput
}

structure EchoInput {
    @httpQuery("abc")
    string: QueryList
}

structure EchoOutput {
    string: QueryList
}

list QueryList{
    member: String
}
```

joewyz:
Current implementation of protocol test provider only validates the generated endpoint and fails to validate the query parameters. 

This PR contains:
1. Add validation for the query parameters in protocol tests if `params` is not empty.
2. Fix the failure of serializing list member with `@httpQuery` (this should also resolve #937)
3. Fix the precedence issue when both `@httpQuery` and `@httpQueryParams` assign same query param.
4. Fix the failure of including hardcoded query params from `@http` trait's `uri` field.
5. Fix the failure of generating methods `outputStreamMember()`, `inputStreamMember()` and `idempotencyTokenMember()` in operation generator.
6. Add new context key in `InjectIdempotencyTokenPlugin` to specify token provider for protocol tests.
7. Clean up some protocol tests for RpcV2Cbor.

### DIFF
--- a/http/http-binding/src/main/java/software/amazon/smithy/java/http/binding/HttpQuerySerializer.java
+++ b/http/http-binding/src/main/java/software/amazon/smithy/java/http/binding/HttpQuerySerializer.java
@@ -7,7 +7,6 @@ package software.amazon.smithy.java.http.binding;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
-import java.nio.ByteBuffer;
 import java.time.Instant;
 import java.util.function.BiConsumer;
 import software.amazon.smithy.java.core.schema.Schema;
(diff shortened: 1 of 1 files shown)


http/http-binding/src/main/java/software/amazon/smithy/java/http/binding/HttpQuerySerializer.java
```diff
@@ -134,4 +134,75 @@ public void writeTimestamp(Schema schema, Instant value) {
...
+
+        @Override
+        public void writeString(Schema schema, String value) {
+            writeQuery(parentTrait, value);
+        }
+
+        @Override
+        public void writeBlob(Schema schema, ByteBuffer value) {
+            writeQuery(parentTrait, ByteBufferUtils.base64Encode(value));
+        }
```
sugmanue: Let's remove this, blobs are not allowed for query params, see [here](https://smithy.io/2.0/spec/http-bindings.html#httpquery-trait)

sugmanue (approved):

PR #995, opened 2025-12-15
Add examples trait support

JordonPhillips:
This adds support for generating examples based on the examples trait. It also adds a general ability to generate code that initializes shapes based on a `Node` value as translated to Java code.

Note that the deserialize interface isn't used in this because the intent is to generate something that's naturally readable. Writing out the node and calling deserilize on it wouldn't be very readable, and wouldn't work on things that don't have the deserialize method.

### DIFF
--- a/codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/generators/SnippetGenerator.java
+++ b/codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/generators/SnippetGenerator.java
@@ -11,7 +11,6 @@ import static software.amazon.smithy.java.core.serde.TimestampFormatter.Prelude.
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
-import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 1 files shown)


codegen/plugins/client-codegen/src/test/java/software/amazon/smithy/java/codegen/client/integrations/javadoc/ClientJavadocExamplesIntegrationTest.java
```diff
@@ -0,0 +1,63 @@
...
+        return TEST_FILE;
+    }
+
+    @Test
+    void includesExamples() {
+        var fileContents = getFileStringForClass("client/TestServiceClient");
+        var expected = """
+                     * <h4>Examples</h4>
+                     * <h5>Basic Example</h5>
+                     * <pre>
```
mtdowling: you can use `{@snippet :` now

mtdowling (approved):

PR #1025, opened 2026-01-28
Support SSE + streaming with MCP notifications.

billyhickamzn:
The proxy implementations do not currently support SSE formatting. 

*Description of changes:*

- Added support for reading streamed HTTP responses with SSE formatting (e.g. data: prefix). Responses are read from the buffer, line by line and notifications processed.
   - Relevant specification documentation: https://modelcontextprotocol.io/specification/2025-03-26/basic/transports

- Notifications are passed to a new notification listener for JsonRPCRequest. The current one used JsonRPCRequest. 

- When a notification/list-changed response is received, we invalidate the local cache and make sure the proxies to do a full refresh the next time list/tools is called.

- If an MCP server includes Mcp-Session-Id in its response we cache this and provide it forward for subsequent requests. This is the spec for this https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management.

Tests pass and run locally to ensure proxy behaviour works as expected.

### DIFF
--- a/mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/HttpMcpProxy.java
+++ b/mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/HttpMcpProxy.java
@@ -8,7 +8,6 @@ package software.amazon.smithy.java.mcp.server;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
-import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import software.amazon.smithy.java.auth.api.Signer;
 import software.amazon.smithy.java.client.core.ClientTransport;
@@ -213,7 +212,6 @@ public final class HttpMcpProxy extends McpServerProxy {
                         } else {
                             // This is a response - convert Document to JsonRpcResponse
                             finalResponse = jsonDocument.asShape(JsonRpcResponse.builder());
-                            finalResponse = message;
                         }
                     } catch (Exception e) {
                         LOG.warn("Failed to parse SSE message: {}", jsonData, e);
@@ -297,7 +295,6 @@ public final class HttpMcpProxy extends McpServerProxy {
                 return false;
             }
 
-            
             // If it has a "method" field but no "id", it's a notification
             return doc.getMember("id") == null && doc.getMember("method") != null;
         } catch (Exception e) {


mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/HttpMcpProxy.java
```diff
@@ -137,6 +164,101 @@ public CompletableFuture<JsonRpcResponse> rpc(JsonRpcRequest request) {
...
+    private JsonRpcResponse parseSseResponse(HttpResponse response, JsonRpcRequest request) {
+        try {
+            byte[] bodyBytes = ByteBufferUtils.getBytes(response.body().asByteBuffer());
+            String sseContent = new String(bodyBytes, StandardCharsets.UTF_8);
+
+            JsonRpcResponse finalResponse = null;
+            String[] lines = sseContent.split("\n", -1); // Use -1 to include trailing empty strings
+            StringBuilder dataBuffer = new StringBuilder();
+
+            for (String line : lines) {
```
rhernandez35: You can avoid an intermediate `String[]` by using `String::lines`, for example:

```
Iterable<String> lines = sseContent.lines()::iterator;
for (var line : lines) { ... }
```

`sseConte.lines().forEach(line -> ...)` also works if you're ok with a state-capturing lambda here (though I would personally prefer the foreach loop)

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/HttpMcpProxy.java
```diff
@@ -125,10 +142,23 @@ public CompletableFuture<JsonRpcResponse> rpc(JsonRpcRequest request) {
             HttpResponse response = transport.send(context, httpRequest);
             LOG.trace("Received HTTP response with status: {}", response.statusCode());
 
+            // Extract and store session ID from response
+            String responseSessionId = response.headers().firstValue("Mcp-Session-Id");
+            if (responseSessionId != null) {
+                sessionId = responseSessionId;
+                LOG.debug("Stored session ID from response: {}", responseSessionId);
+            }
```
adwsingh: This is only allowed during `initialize` but we are accepting it during any responses.

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/HttpMcpProxy.java
```diff
@@ -137,6 +167,134 @@ public CompletableFuture<JsonRpcResponse> rpc(JsonRpcRequest request) {
...
+                        continue;
+                    }
+
+                    try {
+                        // Check if it's a notification by checking for top-level "id" field
+                        // Notifications have "method" but no "id", responses have "id"
+                        if (isNotification(jsonData)) {
+                            // This is a notification - parse as JsonRpcRequest and forward
+                            JsonRpcRequest notification =
+                                    JSON_CODEC.deserializeShape(jsonData, JsonRpcRequest.builder());
```
adwsingh: We are double parsing the JSON here. You can convert the Document to JsonRpcRequest.

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpServerProxy.java
```diff
@@ -118,5 +121,28 @@ protected void notify(JsonRpcResponse response) {
...
+        }
+    }
+
+    /**
+     * Updates the request notification consumer for this proxy.
+     * This allows forwarding notifications that have a method field.
+     */
+    public void updateRequestNotificationConsumer(Consumer<JsonRpcRequest> consumer) {
+        requestNotificationConsumer.set(consumer);
+    }
```
adwsingh: We don't need a separate method for this, take this as a part of `initialize` method.

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpService.java
```diff
@@ -386,7 +415,23 @@ public void addNewService(String id, Service service) {
...
+    }
+
+    /**
+     * Adds a new proxy and initializes it with optional request notification writer.
+     */
+    public void addNewProxy(
+            McpServerProxy mcpServerProxy,
+            Consumer<JsonRpcResponse> responseWriter,
+            Consumer<JsonRpcRequest> notificationWriter
+    ) {
```
adwsingh: I don't think we need a new method, just modify the existing method and change callers.

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpService.java
```diff
@@ -412,6 +457,27 @@ public void addNewProxy(McpServerProxy mcpServerProxy, Consumer<JsonRpcResponse>
...
+
+        // Re-fetch tools from all proxies
+        for (McpServerProxy proxy : proxies.values()) {
+            try {
+                List<ToolInfo> proxyTools = proxy.listTools();
+                for (var toolInfo : proxyTools) {
+                    tools.put(toolInfo.getName(), new Tool(toolInfo, proxy.name(), proxy));
+                }
+            } catch (Exception e) {
+                LOG.error("Failed to refresh tools from proxy: " + proxy.name(), e);
```
adwsingh: Don't silently swallow. Don't catch the exception and let it bubble up.

adwsingh (changes requested):

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/HttpMcpProxy.java
```diff
@@ -137,6 +176,145 @@ public CompletableFuture<JsonRpcResponse> rpc(JsonRpcRequest request) {
...
+            byte[] bodyBytes = ByteBufferUtils.getBytes(response.body().asByteBuffer());
+            String sseContent = new String(bodyBytes, StandardCharsets.UTF_8);
+
+            JsonRpcResponse finalResponse = null;
+            Iterable<String> lines = sseContent.lines()::iterator;
+            StringBuilder dataBuffer = new StringBuilder();
+
+            for (String line : lines) {
+                if (line.startsWith("data:")) {
+                    dataBuffer.append(line.substring(5).replaceFirst("^ ", ""));
```
adwsingh: `replaceFirst()` compiles a regex pattern on every call.
```suggestion
                    var value = line.substring(5);
                    dataBuffer.append(value.startsWith(" ") ? value.substring(1) : value);
```

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/HttpMcpProxy.java
```diff
@@ -137,6 +176,145 @@ public CompletableFuture<JsonRpcResponse> rpc(JsonRpcRequest request) {
...
+                        Document jsonDocument = JSON_CODEC.createDeserializer(jsonData.getBytes(StandardCharsets.UTF_8))
+                                .readDocument();
+
+                        // Check if it's a notification by checking for top-level "id" field
+                        // Notifications have "method" but no "id", responses have "id"
+                        if (isNotification(jsonDocument)) {
+                            // This is a notification - convert Document to JsonRpcRequest and forward
+                            JsonRpcRequest notification = JsonRpcRequest.builder()
+                                    .deserialize(jsonDocument.createDeserializer())
+                                    .build();
```
adwsingh: You can just do `.asShape`.
```suggestion
                            JsonRpcRequest notification = jsonDocument.asShape(JsonRpcRequest.builder());
```

adwsingh (changes requested):

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/HttpMcpProxy.java
```diff
@@ -38,6 +41,7 @@ public final class HttpMcpProxy extends McpServerProxy {
     private final String name;
     private final Signer<HttpRequest, ?> signer;
     private final Duration timeout;
+    private volatile String sessionId;
```
rhernandez35: We don't read this field outside of null-checking it on line 156. Is that intentional or are we supposed to be including this in outbound proxied calls after initialization?

rhernandez35: Yep, I missed that

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/HttpMcpProxy.java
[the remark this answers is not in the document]
adwsingh: I took a different approach. We can't use `Charater.isWhitespace` because  the SSE spec only allows a `U+0020` space. https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream

adwsingh (approved):

PR #1029, opened 2026-02-04
Add support for AWS Query protocol in the client

adwsingh:
*Issue #, if available:*

*Description of changes:*


aws/client/aws-client-awsquery/src/main/java/software/amazon/smithy/java/aws/client/awsquery/AwsQueryClientProtocol.java
```diff
@@ -0,0 +1,178 @@
...
+        if (!Unit.ID.equals(operation.inputSchema().id())) {
+            input.serializeMembers(serializer);
+        }
+
+        ByteBuffer body = serializer.finish();
+
+        return HttpRequest.builder()
+                .method("POST")
+                .uri(endpoint)
+                .headers(HttpHeaders.of(Map.of("Content-Type", CONTENT_TYPE_LIST)))
```
mtdowling: These are immutable headers, so they could be shared across all created requests using a static

PR #1035, opened 2026-02-16
Implement new event streams API

sugmanue:
*Issue #, if available:*

*Description of changes:*

Implements a new event streams API that's VT friendly by blocking when the caller reads and writes events. 


core/src/main/java/software/amazon/smithy/java/core/serde/event/DefaultEventStreamWriter.java
```diff
@@ -0,0 +1,237 @@
...
+    /**
+     * Closes the stream with an error, signaling to the consumer that an error occurred.
+     *
+     * @param e the error that occurred (must not be null)
+     */
+    @Override
+    public void closeWithError(Exception e) {
+        Objects.requireNonNull(e, "exception");
+        if (!closed) {
+            closed = true;
```
rhernandez35: nonatomic get-and-set of a volatile variable is a code smell. either use a compare-and-set primitive via AtomicIntegerFieldUpdater or AtomicBoolean, or make the field non-volatile.

core/src/main/java/software/amazon/smithy/java/core/serde/event/DefaultEventStreamWriter.java
```diff
@@ -0,0 +1,237 @@
...
+            Thread.currentThread().interrupt();
+            throw new RuntimeException("Interrupted while writing event", e);
+        }
+    }
+
+    @Override
+    public void bootstrap(Bootstrap<IE, F> bootstrap) {
+        if (readyLatch.getCount() == 0) {
+            throw new IllegalStateException("bootstrap has been already called");
+        }
```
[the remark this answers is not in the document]
sugmanue: This is there just to catch an bugs from the calling code to call it twice. We don't expect any concurrent callers for this code.

http/http-binding/src/main/java/software/amazon/smithy/java/http/binding/RequestSerializer.java
```diff
@@ -165,14 +165,21 @@ public HttpRequest serializeRequest() {
                 .method(httpTrait.getMethod())
                 .uri(targetEndpoint);
 
-        var eventStream = (Flow.Publisher<SerializableStruct>) serializer.getEventStream();
+        var eventStream = serializer.getEventStream();
         if (eventStream != null && operation instanceof InputEventStreamingApiOperation<?, ?, ?>) {
-            builder.body(EventStreamFrameEncodingProcessor.create(eventStream, eventStreamEncodingFactory));
+            ProtocolEventStreamWriter<SerializableStruct, SerializableStruct, Frame<?>> writer =
+                    ProtocolEventStreamWriter.toInternal(eventStream);
```
mtdowling: toInternal seems like a leftover from the previous naming. Just call it `of`: `ProtocolEventStreamWriter.of`

core/src/main/java/software/amazon/smithy/java/core/serde/event/EventEncoderFactory.java
```diff
@@ -5,6 +5,7 @@
 
 package software.amazon.smithy.java.core.serde.event;
 
+// TODO javadoc
```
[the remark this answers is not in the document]
sugmanue: Added the javadoc to both, the encoder and decoder factories interfaces.

core/src/main/java/software/amazon/smithy/java/core/serde/event/DefaultEventStreamWriter.java
```diff
@@ -29,6 +31,10 @@ final class DefaultEventStreamWriter<IE extends SerializableStruct, T extends Se
         F extends Frame<?>>
         implements ProtocolEventStreamWriter<T, IE, F> {
     private static final InternalLogger LOGGER = InternalLogger.getLogger(DefaultEventStreamWriter.class);
+    /**
+     * Default timeout to block waiting to write.
+     */
+    private static final int WRITE_TIMEOUT_MILLIS = 1500;
```
mtdowling: Main concern here is -- is this universally enough time? I am not positive 1.5 seconds is enough. Not sure if this has to be configurable _yet_, but maybe this should be more like 10 seconds or something like that to account for long setup, but still give _some_ mitigation against hanging forever.

mtdowling (approved):

PR #1040, opened 2026-02-25
Send back a notification after successful initialize

adwsingh:
*Issue #, if available:*

*Description of changes:*

Send back a notification after successful initialize


adwsingh:
> Any particular reason for doing this? I thought silent accept was good enough.

Yeah, the MCP spec mandates that the client send back a initialized notification.

See, https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle

> After successful initialization, the client MUST send an initialized notification to indicate it is ready to begin normal operations:

PR #1064, opened 2026-03-12
Properly handle event payload members of blob and string type

sugmanue:
*Issue #, if available:*

*Description of changes:*

Properly handle event payload members of blob and string type, those are serialized directly into the event payload using a different content-type header. 

Serializes a member with the trait `@eventPayload` directly into the payload of the event. The following member types have special handling:
* **blob** For blob members the non-encoded bytes are the payload of the event and the event `:content-type` header is set to "application/octet-stream"
* **string** For string members the UTF-8 encoded bytes are the payload of the event and the event `:content-type` header is set to "text/plain"
* If the payload is empty, no `:content-type` is added as header of the message
* Any other is encoded with the underlying protocol and uses the content type defined for it, unless 


aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/AwsEventShapeEncoder.java
```diff
@@ -106,146 +202,187 @@ private boolean isInitialRequest(Schema schema) {
...
-                    ":error-code",
-                    HeaderValue.fromString(es.getErrorCode()),
-                    ":error-message",
-                    HeaderValue.fromString(es.getMessage()));
-            frame = new AwsEventFrame(new Message(headers, new byte[0]));
+    /**
+     * Returns true if the schema have any members that ought to be serialized in
+     * the event payload.
+     */
+    static boolean hasPayloadMembers(Schema struct) {
```
rhernandez35: this isn't something to fix in this PR, but this feels like the kind of thing we'd want to attach as a trait on the struct

aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/AwsEventShapeEncoder.java
```diff
@@ -106,146 +202,187 @@ private boolean isInitialRequest(Schema schema) {
...
+     * Serializes an event, class implements the logic to serialize into the headers,
+     * or the payload of the message depending on the presence of the event serialization
+     * traits.
+     */
     static class EventSerializer extends SpecificShapeSerializer {
+        private final OutputStream out;
+        private final Codec codec;
         private final EventHeaderSerializer headerSerializer;
-        private final ShapeSerializer baseSerializer;
+        private final AtomicReference<String> contentTypeHolder;
```
rhernandez35: like above, AtomicReference is the wrong thing to use if we don't need atomicity. it's not appropriate to use as a generic mutable reference, unless you commit to only using `getPlain` and `setPlain` to interact with it (but even that would be an abuse of its APIs)

examples/transcribestreaming-client/README.md
```diff
@@ -0,0 +1,18 @@
+## Example: Transcribe Streaming Client
```
rhernandez35: nice

aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/AwsEventShapeEncoder.java
sugmanue: We could pre-compute and cache using a set of schemas, I will look into it in a follow up PR.

PR #1075, opened 2026-03-13
Add virtual-thread-based HTTP client

mtdowling:
Adds a blocking HTTP client built for virtual threads with an HTTP/1.1 and HTTP/2 implementations, connection pooling, flow control, HPACK encoding/decoding, and comprehensive test coverage including fuzz tests.

*Issue #, if available:*

*Description of changes:*


http/http-client/src/main/java/software/amazon/smithy/java/http/client/RequestOptions.java
```diff
@@ -0,0 +1,191 @@
...
+         */
+        public RequestOptions build() {
+            // Take-and-replace to avoid defensive copies
+            Context ctx = context != null ? context : Context.create();
+            context = null;
+
+            List<HttpInterceptor> ints = interceptors != null ? interceptors : List.of();
+            interceptors = null;
+
+            return new RequestOptions(ctx, requestTimeout, ints);
```
sugmanue: Nit, I think is better to have a clear expected state after build and clear all the references that can be cleared.

```suggestion

            Duration reqTimeout = requestTimeout;
            requestTimeout = null;

            return new RequestOptions(ctx, reqTimeout, ints);
```

http/http-client/src/main/java/software/amazon/smithy/java/http/client/HttpClient.java
```diff
@@ -0,0 +1,252 @@
...
+         * @param interceptor the interceptor to add
+         * @return this builder
+         */
+        public Builder addInterceptor(HttpInterceptor interceptor) {
+            interceptors.add(Objects.requireNonNull(interceptor, "interceptor"));
+            return this;
+        }
+
+        /**
+         * Add an interceptor to the front of the list of interceptors ot apply.
```
sugmanue:
```suggestion
         * Add an interceptor to the front of the list of interceptors to apply.
```

http/http-client/src/main/java/software/amazon/smithy/java/http/client/BufferedHttpExchange.java
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * <p>Used when an interceptor short-circuits the request via {@code handleRequest()}
+ * or {@code onError()}, returning a pre-existing response (e.g., from cache).
+ *
+ * <p>The request body is a no-op since the request was never actually sent.
+ */
+final class BufferedHttpExchange implements HttpExchange {
+    private final HttpRequest request;
+    private final HttpResponse response;
+    private final OutputStream noopRequestBody = OutputStream.nullOutputStream();
```
sugmanue: Nit, this can be static

http/http-client/src/main/java/software/amazon/smithy/java/http/client/connection/H2ConnectionManager.java
```diff
@@ -0,0 +1,368 @@
...
+
+    private static final H2Connection[] EMPTY = new H2Connection[0];
+
+    // Soft limit as a fraction of streamsPerConnection. When all connections exceed this threshold,
+    // we try to create a new connection (if under max).
+    private static final int DEFAULT_SOFT_LIMIT_DIVISOR = 4;
+    private static final int DEFAULT_SOFT_LIMIT_FLOOR = 25;
+
+    private final ConcurrentHashMap<Route, RouteState> routes = new ConcurrentHashMap<>();
+    private final int streamsPerConnection;
```
sugmanue: Nit, 'streamsPerConnection' is assigned but never used. The cap on streams per connections is done is `H2ConnectionManager` and `WatermarkLoadBalancer`.

http/http-client/src/main/java/software/amazon/smithy/java/http/client/connection/H2ConnectionManager.java
```diff
@@ -0,0 +1,368 @@
...
+                LOGGER.debug("All {} connections saturated for route {}, waiting for capacity "
+                        + "(canExpand={}, selected={})", connCount, route, canExpand, selected);
+                long remainingNanos = deadlineNanos - System.nanoTime();
+                if (remainingNanos <= 0) {
+                    throw new IOException("Acquire timeout: no connection available after "
+                            + acquireTimeoutMs + "ms for " + route);
+                }
+
+                try {
+                    state.available.awaitNanos(remainingNanos);
```
sugmanue, line 165: Looks like `state.available` is not signaled when a stream is closed which will make room for a new one.

http/http-client/src/main/java/software/amazon/smithy/java/http/client/connection/HttpVersionPolicy.java
```diff
@@ -0,0 +1,49 @@
...
+    /** HTTP/2 over TLS only. Negotiates only "h2" via ALPN. Fails if server doesn't support. */
+    ENFORCE_HTTP_2(new String[] {"h2"}),
+
+    /** Prefer HTTP/2, fall back to HTTP/1.1. Uses HTTP/1.1 for cleartext. Recommended default. */
+    AUTOMATIC(new String[] {"h2", "http/1.1"}),
+
+    /** HTTP/2 over cleartext (h2c) using prior knowledge. */
+    H2C_PRIOR_KNOWLEDGE(new String[] {"h2"});
+
+    private final String[] alpnProtocols;
```
sugmanue, line 24: Nit, we return this value which can be mutated by the caller, probably low risk but still fishy.

http/http-client/src/main/java/software/amazon/smithy/java/http/client/connection/HttpConnectionPool.java
```diff
@@ -0,0 +1,551 @@
...
+        this.versionPolicy = builder.versionPolicy;
+        DnsResolver dnsResolver = builder.dnsResolver != null ? builder.dnsResolver : DnsResolver.system();
+        SSLContext sslContext = builder.sslContext;
+
+        if (sslContext == null) {
+            try {
+                sslContext = SSLContext.getDefault();
+            } catch (NoSuchAlgorithmException e) {
+                throw new IllegalStateException("No default SSLContext available", e);
+            }
```
sugmanue: `sslContext` is already resolved in the builder.

http/http-client/src/main/java/software/amazon/smithy/java/http/client/h1/H1Exchange.java
```diff
@@ -0,0 +1,578 @@
...
+                // Parse as final response, must not send body
+                parseStatusAndHeaders(code, in);
+                requestWritten = true; // Skip body transmission
+                throw new IOException("Server sent final response " + code
+                        + " before request body; body must not be written");
+            }
+        } catch (SocketTimeoutException e) {
+            // Timeout waiting for 100 Continue - proceed with body anyway
+            // Some servers don't support 100-continue and just ignore it
+            return;
```
sugmanue: Nit, the `return` can be removed.

http/http-client/src/main/java/software/amazon/smithy/java/http/client/h1/H1Exchange.java
```diff
@@ -0,0 +1,578 @@
...
+
+            // Check Connection header to determine keep-alive behavior
+            // HTTP/1.1 defaults to keep-alive, HTTP/1.0 defaults to close
+            if ("connection".equals(name)) {
+                String value = headers.firstValue(name);
+                if ("close".equalsIgnoreCase(value)) {
+                    sawConnectionClose = true;
+                } else if ("keep-alive".equalsIgnoreCase(value)) {
+                    // Explicit keep-alive (needed for HTTP/1.0)
+                    connection.setKeepAlive(true);
```
sugmanue: Is the idea of keeping the flag `sawConnectionClose` and calling `setKeepAlive(false)` outside the loop to give it higher priority if two connection headers show up? It's not super clear why for `keep-alive` the setter is called here and for `close` is called outside the loop.

http/http-client/src/main/java/software/amazon/smithy/java/http/client/h1/H1Exchange.java
```diff
@@ -0,0 +1,578 @@
...
+    public void close() throws IOException {
+        if (!closed) {
+            closed = true;
+            if (responseIn != null) {
+                responseIn.close();
+            }
+            if (requestOut != null) {
+                requestOut.close();
+            }
+            connection.releaseExchange();
```
sugmanue: Nit, the input and output streams can't throw but this will make the intent clear. A commend would as well.
```suggestion
            try {
                if (responseIn != null) {
                    responseIn.close();
                }
                if (requestOut != null) {
                    requestOut.close();
                }
            } finally {
                connection.releaseExchange();
            }
```

http/http-client/src/main/java/software/amazon/smithy/java/http/client/h1/ChunkedInputStream.java
```diff
@@ -0,0 +1,291 @@
...
+    private boolean closed;
+    private final byte[] lineBuffer = new byte[MAX_LINE_LENGTH];
+    private HttpHeaders trailers; // Trailer headers parsed from final chunk (RFC 7230 Section 4.1.2)
+
+    ChunkedInputStream(UnsyncBufferedInputStream delegate) {
+        this.delegate = delegate;
+    }
+
+    private static long readMaxChunkSize() {
+        String property = System.getProperty("SMITHY_HTTP_CLIENT_MAX_CHUNK_SIZE");
```
sugmanue: Nit, I think we should use a more java property friendly name with lowercase and dots to separate segments, e.g., `smithy.httpClient.maxChunkSize` or similar

http/http-client/src/main/java/software/amazon/smithy/java/http/client/h2/FlowControlWindow.java
```diff
@@ -0,0 +1,147 @@
...
+            // Slow path: poll with short intervals to avoid timed-wait contention
+            long remainingNs = TimeUnit.MILLISECONDS.toNanos(timeoutMs);
+            while (window <= 0) {
+                if (remainingNs <= 0) {
+                    return 0; // Timeout
+                }
+                // Use short poll interval instead of full timeout
+                long waitNs = Math.min(remainingNs, POLL_INTERVAL_NS);
+                remainingNs = available.awaitNanos(waitNs);
+            }
```
sugmanue: `awaitNanos` returns _an estimate of the nanosTimeout value minus the time spent waiting upon return from this method_, if `remainingNs > POLL_INTERVAL_NS` it will be capped to `POLL_INTERVAL_NS`. We can use a deadline as we do in `closeExchanges`.
```suggestion
            long deadlineNs = System.nanoTime() + TimeUnit.MILLISECONDS.toNanos(timeoutMs);
            while (window <= 0) {
                long remainingNs = deadlineNs - System.nanoTime();
                if (remainingNs <= 0) {
                    return 0; // Timeout
                }
                // Use short poll interval instead of full timeout
                available.awaitNanos(Math.min(remainingNs, POLL_INTERVAL_NS));
            }
```

http/http-client/src/main/java/software/amazon/smithy/java/http/client/h2/H2Muxer.java
```diff
@@ -0,0 +1,830 @@
...
+        workQueue.add(item);
+        signalWriter();
+    }
+
+    void queueControlFrame(int streamId, ControlFrameType frameType, Object payload, long timeoutMs) {
+        H2MuxerWorkItem item = switch (frameType) {
+            case RST_STREAM -> new H2MuxerWorkItem.WriteRst(streamId, (Integer) payload);
+            case WINDOW_UPDATE -> new H2MuxerWorkItem.WriteWindowUpdate(streamId, (Integer) payload);
+            case SETTINGS_ACK -> SETTINGS_ACK;
+            case PING -> new H2MuxerWorkItem.WritePing((byte[]) payload, false);
```
sugmanue: We are queuing a PING response, therefore we need to ACK

> When set, the ACK flag indicates that this PING frame is a PING response.

```suggestion
            case PING -> new H2MuxerWorkItem.WritePing((byte[]) payload, true);
```

http/http-client/src/main/java/software/amazon/smithy/java/http/client/h2/H2Connection.java
```diff
@@ -0,0 +1,808 @@
...
+
+    // ==================== Reader Thread ====================
+
+    // Track last stream for batched signaling (stream-switch detection).
+    // With lock-free signaling, flushing the previous stream is cheap (just LockSupport.unpark).
+    private H2Exchange lastDataExchange;
+
+    private void readerLoop() {
+        try {
+            while (state.get() == State.CONNECTED) {
```
sugmanue: `handleGoaway()` sets state to `SHUTTING_DOWN` but streams `<= lastStreamId` should be allowed to complete, I think we should do instead.
```suggestion
            while (state.get() != State.CLOSED) {
```

sugmanue (changes requested):

http/http-client/src/main/java/software/amazon/smithy/java/http/client/connection/H2ConnectionManager.java
```diff
@@ -0,0 +1,368 @@
...
+                boolean canExpand = totalConns < maxConnectionsForRoute;
+                int selected = loadBalancer.select(state.activeStreamsBuf,
+                        connCount,
+                        canExpand ? maxConnectionsForRoute : connCount);
+
+                if (selected >= 0) {
+                    notifyAcquire(snapshot[selected], true);
+                    return snapshot[selected];
+                } else if (selected == H2LoadBalancer.CREATE_NEW && canExpand) {
+                    state.pendingCreations++;
```
sugmanue, line 151: Another one found while testing, avoid creating two connections with one stream each instead of just one with two streams.
```suggestion
                    if (state.pendingCreations > 0) {
                        // Another thread is already creating a connection, wait for it
                        // instead of creating a redundant one.
                        long remainingNanos = deadlineNanos - System.nanoTime();
                        if (remainingNanos <= 0) {
                            throw new IOException("Acquire timeout: no connection available after "
                                    + acquireTimeoutMs + "ms for " + route);
                        }
                        try {
                            state.available.awaitNanos(remainingNanos);
                        } catch (InterruptedException e) {
                            Thread.currentThread().interrupt();
                            throw new IOException("Interrupted while waiting for connection", e);
                        }
                        continue;
                    }
                    state.pendingCreations++;
```

http/http-client/src/main/java/software/amazon/smithy/java/http/client/h1/H1Exchange.java
mtdowling: Good call. I refactored to defer both and resolve after the loop with a single setKeepAlive call

http/http-client/src/main/java/software/amazon/smithy/java/http/client/h2/FlowControlWindow.java
mtdowling: switched to a System.nanoTime() - was using the ticks style optimization the rest of the code base uses, but this isn't a hot path that needs it.

sugmanue (approved):

sugmanue:
😭

Issue #1082, opened 2026-03-17
java-codegen types-only mode crashes with ExpectationNotMetException on models using @mixin shapes


AgentKWang:
## Description

The unified `java-codegen` plugin (introduced in #1057) crashes in types-only mode (`"modes": ["types"]`) when the model contains `@mixin` shapes.

```
ExpectationNotMetException: Shape not found in model: com.example#MyMixin
    at software.amazon.smithy.model.knowledge.EventStreamIndex.computeEvents(EventStreamIndex.java:54)
    at software.amazon.smithy.java.codegen.generators.OperationGenerator.accept(OperationGenerator.java:55)
    at software.amazon.smithy.java.codegen.DirectedJavaCodegen.generateOperation(DirectedJavaCodegen.java:136)
    at software.amazon.smithy.java.codegen.JavaCodegenPlugin.executeTypesMode(JavaCodegenPlugin.java:115)
```

## Root Cause

In types-only mode, `executeTypesMode()` calls `SyntheticServiceTransform.transform()` which builds a synthetic model via `ModelTransformer.replaceShapes()`. This model has an incomplete transitive closure — `@mixin` shapes are flattened during model assembly, and `EventStreamIndex` calls `model.expectShape()` on member targets that reference these flattened mixin shapes, which are not resolvable in the transformed model.

`DirectedJavaCodegen.generateOperation()` runs `OperationGenerator` unconditionally, including in types-only mode. `OperationGenerator` invokes `EventStreamIndex.of(model)` which triggers the crash. The old `DirectedJavaTypeCodegen` (removed in #1057) never implemented `generateOperation()`, so this code path was never hit before.

## Reproduction

`smithy-build.json`:

```json
{
  "version": "1.0",
  "plugins": {
    "java-codegen": {
      "namespace": "com.example",
      "selector": "[id|namespace ^= 'com.example']",
      "modes": ["types"]
    }
  }
}
```

Model with a `@mixin` used by an error structure:

```smithy
$version: "2.0"

namespace com.example

@mixin
structure BaseError {
    @required
    message: String
}

@error("server")
@httpError(500)
structure InternalException with [BaseError] {}

@http(code: 200, method: "POST", uri: "/doSomething")
operation DoSomething {
    input: DoSomethingInput
    output: DoSomethingOutput
    errors: [InternalException]
}

structure DoSomethingInput {}
structure DoSomethingOutput {}

service MyService {
    version: "1.0"
    operations: [DoSomething]
}
[...]

PR #1083, opened 2026-03-17
Skip OperationGenerator in types-only codegen mode

AgentKWang:
*Issue #, if available:* #1082

*Description of changes:*
In types-only mode, OperationGenerator is invoked for operations added to the synthetic service by SyntheticServiceTransform. OperationGenerator calls EventStreamIndex, which iterates all operation input/output members via model.expectShape(). The synthetic model built by SyntheticServiceTransform has an incomplete transitive closure — @mixin shapes (e.g. BaseError, ComponentInstanceMixin) are flattened during model assembly and not resolvable via expectShape() in the transformed model, causing ExpectationNotMetException.

The old DirectedJavaTypeCodegen (removed in #1057) never implemented generateOperation(), so this path was never hit. The unified DirectedJavaCodegen runs OperationGenerator unconditionally, which is unnecessary in types-only mode — operation input/output types are generated through generateStructure(), not generateOperation().

This adds an isTypesOnly() guard to generateOperation() and refactors generateService() to use the same helper.


codegen/codegen-plugin/src/main/java/software/amazon/smithy/java/codegen/DirectedJavaCodegen.java
```diff
@@ -130,6 +130,13 @@ public void generateOperation(GenerateOperationDirective<CodeGenerationContext,
         if (isSynthetic(directive.shape())) {
             return;
         }
+        // In types-only mode, skip OperationGenerator. Operation input/output types are still
+        // generated via generateStructure(). OperationGenerator creates operation wiring classes
+        // and invokes EventStreamIndex, which fails on models built by SyntheticServiceTransform
+        // when @mixin shapes are present (incomplete transitive closure).
```
sugmanue: For types only it makes perfect sense to skip operations, I think is better to remove this in the same way we don't have an explanation as of why we skip if the shape is synthetic.
```suggestion
```

adwsingh (approved):

PR #1142, opened 2026-04-06
Extend McpMetricsObserver with completion, error, and tools/list callbacks

psp65:
## Summary
  - Add three new default methods to `McpMetricsObserver`: `onToolCallComplete`, `onToolCallError`, and `onToolsList`
  - Instrument `McpService.handleToolsCall()` with `System.nanoTime()` latency tracking on both local and proxy code paths
  - Instrument `McpService.handleToolsList()` with tool count callback
  - Every `onToolCall` is now guaranteed a matching `onToolCallComplete`, including the "tool not found" error path

  ## Motivation
  The existing `McpMetricsObserver` only tracks tool call initiation (`onToolCall`) and initialization (`onInitialize`). There is no visibility into:
  - Whether tool calls succeed or fail
  - How long tool calls take (latency)
  - Which server handled the call
  - Whether the call was proxied or handled locally
  - How often clients request the tool list

  These callbacks enable downstream consumers (e.g., CLI telemetry) to emit per-tool counters, per-server breakdowns, latency timings, and error rates.

### DIFF
--- a/mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpMetricsObserver.java
+++ b/mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpMetricsObserver.java
@@ -5,6 +5,7 @@
 
 package software.amazon.smithy.java.mcp.server;
 
+import java.time.Duration;
 import software.amazon.smithy.utils.SmithyUnstableApi;
 
 /**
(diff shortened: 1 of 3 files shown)


mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpMetricsObserver.java
```diff
@@ -33,4 +33,34 @@ void onToolCall(
...
+    /**
+     * Called after a tool call completes (success or failure), with timing and server context.
+     */
+    default void onToolCallComplete(
+            String method,
+            String toolName,
+            String serverId,
+            long latencyNanos,
+            boolean success,
+            boolean isProxy
```
adwsingh, line 62: I think the params warrant java docs, its not super clear to me what each mean.

Also why latencyNanos as Long, instead of Duration?

adwsingh (changes requested):

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpMetricsObserver.java
psp65: Ack, addressed both.

psp65:
Closing in lieu of https://github.com/smithy-lang/smithy-java/pull/1174

PR #1149, opened 2026-04-11
Implement standard and adaptive retry strategies

sugmanue:
*Issue #, if available:*

*Description of changes:*

Implements retries strategies for Smithy-Java.


client/client-core/src/main/java/software/amazon/smithy/java/client/core/Client.java
```diff
@@ -49,12 +51,13 @@ protected Client(Builder<?, ?> builder) {
...
         this.typeRegistry = typeRegistry();
-
         if (config.retryStrategy() != null) {
             this.retryStrategy = config.retryStrategy();
         } else {
-            // TODO: Pick a better default retry strategy.
-            this.retryStrategy = RetryStrategy.noRetries();
+            this.retryStrategy = StandardRetryStrategy.create();
+        }
+        if (retryStrategy instanceof Claimable c) {
```
mtdowling, line 61: Can you add a code comment on why we claim right away?

retries-api/src/main/java/software/amazon/smithy/java/retries/api/AcquireInitialTokenRequest.java
```diff
@@ -8,11 +8,17 @@
...
+ *                      <p>A scope should be a unique string describing the smallest possible scope of failure for the attempts
+ *                      about to be executed. In practical terms, this is a key for the token bucket used to throttle request
+ *                      attempts. All attempts with the same scope share the same token bucket within the same
+ *                      {@link RetryStrategy}, ensuring that token-bucket throttling for requests against one resource do not
+ *                      result in throttling for requests against other, unrelated resources.
+ * @param isLongPolling True if the operation is long polling. Retries strategies can use this information to allow
+ *                      retries even when otherwise the retry will not be allowed.
  */
-public record AcquireInitialTokenRequest(String scope) {}
+public record AcquireInitialTokenRequest(String scope, boolean isLongPolling) {
```
mtdowling: I wonder if it might be better to add a flags mask instead?

retries/src/main/java/software/amazon/smithy/java/retries/Constants.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package software.amazon.smithy.java.retries;
+
+import java.time.Duration;
+
+/**
+ * Constants used by retry strategy implementations.
+ */
+final class Constants {
```
mtdowling, line 13: Could / should this be an enum?

retries/src/main/java/software/amazon/smithy/java/retries/DefaultRetryToken.java
```diff
@@ -0,0 +1,194 @@
...
+        private static final int MAX_FAILURES = 10;
+        private String scope;
+        private boolean isLongPolling;
+        private int attempt = 1;
+        private int capacityAcquired = 0;
+        private int capacityRemaining = 0;
+        private ArrayDeque<Throwable> failures;
+
+        Builder() {
+            this.failures = new ArrayDeque<>();
```
mtdowling: can we defer this until we actually have failures?

retries/src/main/java/software/amazon/smithy/java/retries/TokenBucketStore.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * A store to keep token buckets per scope.
+ */
+final class TokenBucketStore {
```
mtdowling, line 16: comment if it's intended ot be thread safe or not

retries/src/main/java/software/amazon/smithy/java/retries/TokenBucket.java
```diff
@@ -0,0 +1,96 @@
...
+
+        return new AcquireResponse(initialRetryTokens, amountToAcquire, amountToAcquire, newCapacity, false);
+    }
+
+    /**
+     * Release a number of tokens back to this bucket. If this number of tokens would exceed the maximum number of
+     * tokens configured for the bucket, the bucket is instead set to the maximum value and the additional tokens are
+     * discarded.
+     */
+    ReleaseResponse release(int amountToRelease) {
```
mtdowling, line 58: This is all internal, including ReleaseResponse. Given that, maybe instead of creating a new ReleaseResponse each time, you could pass in a mutable ReleaseResponse that gets populated by this method instead.

retries/src/main/java/software/amazon/smithy/java/retries/RateLimiterUpdateResponse.java
```diff
@@ -0,0 +1,14 @@
...
+
+package software.amazon.smithy.java.retries;
+
+/**
+ * The result of a rate limiter update operation.
+ *
+ * @param measuredTxRate The measured transaction rate.
+ * @param fillRate       The current fill rate.
+ */
+record RateLimiterUpdateResponse(double measuredTxRate, double fillRate) {}
```
mtdowling: same idea here -- maybe pass in and populate a mutable version. We don't need internal mutability for these. They're just filling in the role of multi-return values, but doing so via allocations

retries/src/main/java/software/amazon/smithy/java/retries/RateLimiterTokenBucketStore.java
```diff
@@ -0,0 +1,49 @@
...
+package software.amazon.smithy.java.retries;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * A store that manages {@link RateLimiterTokenBucket} instances per scope with LRU eviction.
+ */
```
mtdowling, line 17: comment on if it's meant to be thread safe

retries/src/main/java/software/amazon/smithy/java/retries/RateLimiterAcquireResponse.java
```diff
@@ -0,0 +1,15 @@
...
+package software.amazon.smithy.java.retries;
+
+import java.time.Duration;
+
+/**
+ * The result of a {@link RateLimiterTokenBucket#acquirePermit()} call.
+ *
+ * @param delay The amount of time to wait before making the request.
+ */
+record RateLimiterAcquireResponse(Duration delay) {}
```
mtdowling: this is all internal stuff, so just return Duration. If we ever need to return multiple values, come back to this, or a mutable version of it.

retries/src/main/java/software/amazon/smithy/java/retries/RateLimiterClock.java
```diff
@@ -0,0 +1,22 @@
...
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.smithy.java.retries;
+
+/**
+ * A clock used by the rate limiter to measure elapsed time.
+ */
+interface RateLimiterClock {
```
mtdowling, line 12: `@FunctionalInterface`

retries/src/main/java/software/amazon/smithy/java/retries/RateLimiterTokenBucket.java
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        TransientState toTransient() {
+            return new TransientState(this);
+        }
+
+        /**
+         * Returns the current fill rate.
+         */
+        double fillRate() {
```
mtdowling: If this class isn't used outside the outer class, then ditch the getters and mark the class private. I didn't check though.

retries/src/main/java/software/amazon/smithy/java/retries/RateLimiterTokenBucket.java
```diff
@@ -0,0 +1,310 @@
...
+        var update = consumeState(ts -> ts.updateClientSendingRate(clock, false, count, requestCount));
+        return new RateLimiterUpdateResponse(update.newState.measuredTxRate(), update.newState.fillRate());
+    }
+
+    /**
+     * Similar to {@link #updateState} but used when the caller only cares about the side effects of the {@link
+     * Consumer} but not for the value returned.
+     */
+    private StateUpdate<Void> consumeState(Consumer<TransientState> mutator) {
+        return updateState(ts -> {
```
mtdowling: Make it a biFunction and pass in the upvalue `mutator`

mtdowling: or, just make this a normal function as only true/false in the second arg seem to vary

retries/src/main/java/software/amazon/smithy/java/retries/RateLimiterTokenBucket.java
```diff
@@ -0,0 +1,310 @@
...
+        this.requestCount = new AtomicLong(0);
+    }
+
+    /**
+     * Acquire tokens from the bucket. If the bucket contains enough capacity to satisfy the request, this method will
+     * return in {@link RateLimiterAcquireResponse#delay()} a {@link Duration#ZERO} value, otherwise it will return the
+     * amount of time the callers need to wait until enough tokens are refilled.
+     */
+    RateLimiterAcquireResponse acquirePermit() {
+        var update = updateState(ts -> ts.tokenBucketAcquire(clock, 1.0));
```
mtdowling: same thing here -- seems like some restructuring can avid these closures

retries/src/main/java/software/amazon/smithy/java/retries/BaseRetryStrategy.java
```diff
@@ -0,0 +1,411 @@
...
+     */
+    protected Duration computeInitialBackoff(AcquireInitialTokenRequest request) {
+        return Duration.ZERO;
+    }
+
+    /**
+     * Computes the backoff before a retry using the configured backoff strategy. Extending classes can override this
+     * method to compute a different backoff depending on their logic.
+     */
+    protected Duration computeBackoff(RefreshRetryTokenRequest request, DefaultRetryToken token) {
```
mtdowling, line 142: Our exceptions have RetryInfo on them that give protocol hints. How does that interop with all this?

mtdowling: e.g. RetryInfo also has retry after hints `Duration retryAfter();`

mtdowling (changes requested):

retries/src/main/java/software/amazon/smithy/java/retries/BaseRetryStrategy.java
sugmanue: I think it should be part of the downstream call when we set the `RefreshRetryTokenRequest`. I will refactor client to pass this information from the `RetryInfo` instance as the suggested delay to the refresh call.

retries/src/main/java/software/amazon/smithy/java/retries/RateLimiterTokenBucket.java
sugmanue: I will just have the CAS loop on each method. All of that was added to keep it in a single place but that's no actually needed.

retries/src/main/java/software/amazon/smithy/java/retries/RateLimiterUpdateResponse.java
sugmanue: This is only used for tests, I will refactor to avoid creating a new object for every call.

retries/src/main/java/software/amazon/smithy/java/retries/RateLimiterTokenBucket.java
sugmanue: I will just inline all the CAS loops to avoid all this.

retries/src/main/java/software/amazon/smithy/java/retries/TokenBucketStore.java
sugmanue: It is, I will add a comment.

retries/src/main/java/software/amazon/smithy/java/retries/TokenBucket.java
sugmanue: I doesn't work for this case, the strategy itself calls this method and it can be used by multiple threads.

retries/src/main/java/software/amazon/smithy/java/retries/Constants.java
sugmanue: I think it could, but I don't think it should. This class is only a container for global and local to each strategy constants.

retries-api/src/main/java/software/amazon/smithy/java/retries/api/AcquireInitialTokenRequest.java
sugmanue: Sure, I think that's a good idea. I will make that change.

mtdowling (approved):

PR #1157, opened 2026-04-22
feat(mcp): infer ToolAnnotations from @readonly and @idempotent traits

Garnethil:
Background

   - Adds ToolAnnotations support to the MCP server module, automatically inferring readOnlyHint from @readonly and idempotentHint
    from @idempotent Smithy traits on operations.
   - The MCP spec includes these hints to help LLM clients understand tool behavior. Without them, clients must assume worst-case
   defaults (mutable, non-idempotent).

   Testing

   - Added three test cases to McpServerTest: verifies @readonly emits readOnlyHint: true, @idempotent emits idempotentHint: true,
   and plain operations omit annotations entirely.
   - All existing tests pass with updated tool count.

   Links

   - MCP spec ToolAnnotations: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-03-26/schema.json
    (https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/2025-03-26/schema.json)

   ───────────────────────────────────────

### DIFF
--- a/mcp/mcp-schemas/model/main.smithy
+++ b/mcp/mcp-schemas/model/main.smithy
@@ -96,6 +96,13 @@ structure ToolInfo {
     inputSchema: JsonObjectSchema
 
     outputSchema: JsonObjectSchema
+
+    annotations: ToolAnnotations
+}
+
+structure ToolAnnotations {
+    readOnlyHint: Boolean
+    idempotentHint: Boolean
 }
 
 list ToolInfoList {
--- a/mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpService.java
+++ b/mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpService.java
@@ -56,6 +56,7 @@ import software.amazon.smithy.java.mcp.model.PromptInfo;
 import software.amazon.smithy.java.mcp.model.Prompts;
 import software.amazon.smithy.java.mcp.model.ServerInfo;
 import software.amazon.smithy.java.mcp.model.TextContent;
+import software.amazon.smithy.java.mcp.model.ToolAnnotations;
 import software.amazon.smithy.java.mcp.model.ToolInfo;
 import software.amazon.smithy.java.mcp.model.Tools;
 import software.amazon.smithy.java.server.Operation;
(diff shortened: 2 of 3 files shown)


mcp/mcp-schemas/model/main.smithy
```diff
@@ -96,6 +96,16 @@ structure ToolInfo {
     inputSchema: JsonObjectSchema
 
     outputSchema: JsonObjectSchema
+
+    annotations: ToolAnnotations
+}
+
+structure ToolAnnotations {
```
mtdowling, line 103: I see readOnlyHint and idempotentHint are used, but what about the others?

Garnethil: They are part of the MCP Spec but there is no direct annotation in Smithy to use as inference, left this as future work could added as a separate annotation as documented on the linked issue. I could remove this to keep clean. Working on the rebase.

Garnethil: After thinking about this, choose to remove the unused annotation and leave the PR as minimal as possible.

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpService.java
```diff
@@ -588,13 +589,30 @@ private Map<String, Tool> createTools(Map<String, Service> services) {
                                 operation.getApiOperation().outputSchema(),
                                 new HashSet<>(),
                                 cache))
+                        .annotations(createAnnotations(schema))
```
adwsingh, line 602: You can only do this if the MCP Protocol version is `2025-03-26` or later.

adwsingh (changes requested):

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpService.java
Garnethil: Good point, I missed that this was managed by the client. I have applied the filter the same way as the OutputSchema.

adwsingh (approved):

adwsingh:
Thanks for the contribution @Garnethil !

PR #1163, opened 2026-04-29
JDK HTTP client perf

mtdowling:
Improve JDK HTTP client transport performance

On my M1 Mac mini:

- Mixed H2/TLS GET 1 MiB + PUT 1 MiB, c=10
  - original: 32.767 ops/s at conn=1, 30.575 at conn=3
  - current: 49.357 ops/s at conn=1, 43.678 at conn=3
  - change: about +51% at conn=1, about +43% at conn=3
- Tiny RPC, lower is better
  - original: 1460.365 us/op
  - current: 1069.535 us/op
  - change: about 27% lower mean latency

### DIFF
--- a/client/client-http/src/main/java/software/amazon/smithy/java/client/http/JavaHttpClientSmallBodySubscriber.java
+++ b/client/client-http/src/main/java/software/amazon/smithy/java/client/http/JavaHttpClientSmallBodySubscriber.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.smithy.java.client.http;
+
+import java.net.http.HttpHeaders;
+import java.net.http.HttpResponse;
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.Flow;
+import software.amazon.smithy.java.io.datastream.DataStream;
+
+/**
+ * Small-body subscriber for known-length responses that fit in memory cheaply.
+ *
+ * <p>Allocates one array up front and completes with a replayable {@link DataStream}.
+ */
+final class JavaHttpClientSmallBodySubscriber implements HttpResponse.BodySubscriber<DataStream> {
+    private final String contentType;
+    private final byte[] bytes;
+    private int position;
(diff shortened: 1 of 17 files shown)


client/client-http/src/main/java/software/amazon/smithy/java/client/http/JavaHttpClientSmallBodySubscriber.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * <p>Allocates one array up front and completes with a replayable {@link DataStream}.
+ */
+final class JavaHttpClientSmallBodySubscriber implements HttpResponse.BodySubscriber<DataStream> {
+    private final String contentType;
+    private final byte[] bytes;
+    private int position;
+    private final CompletableFuture<DataStream> body = new CompletableFuture<>();
+
+    JavaHttpClientSmallBodySubscriber(java.net.http.HttpHeaders headers, int contentLength) {
```
sugmanue: Nit
```suggestion
    JavaHttpClientSmallBodySubscriber(HttpHeaders headers, int contentLength) {
```

client/client-http/src/main/java/software/amazon/smithy/java/client/http/JavaHttpClientSmallBodySubscriber.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package software.amazon.smithy.java.client.http;
+
+import java.net.http.HttpResponse;
```
sugmanue, line 9:
```suggestion
import java.net.http.HttpHeaders;
import java.net.http.HttpResponse;
```

http/http-client/src/jmh/java/software/amazon/smithy/java/http/client/BenchmarkSupport.java
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        public void throwIfErrored(String label) {
+            if (firstError == null) {
+                return;
+            }
+            if (firstError instanceof RuntimeException runtimeException) {
+                throw new IllegalStateException(label + " failed with " + errors + " error(s)", runtimeException);
+            }
+            throw new IllegalStateException(label + " failed with " + errors + " error(s)", firstError);
```
sugmanue: Why do we make `RutimeException` special? Looks like the result would be the same.

sugmanue (approved):

PR #1166, opened 2026-05-01
Add service schema to ProtocolSettings for service-level traits

sugmanue:
### Summary

When a service has the `@xmlNamespace` trait, the namespace must be applied to the top-level XML element during serialization and deserialization.
The restXml protocol was not handling this because the `XmlCodec` only read `@xmlNamespace` from individual shapes, not from the service.

### Approach

Add `serviceSchema(Schema)` to `ProtocolSettings`. Since `Schema` already carries traits throughout the codebase, protocols can extract whatever service-level traits they need without adding protocol-specific fields to the shared settings class. The `RestXmlClientProtocol` factory reads the `xmlNamespace~ trait from the service schema and passes it to the `XmlCodec` as a default namespace for top-level elements.

### Additional fixes

- `XmlDeserializer`: Namespace-prefixed attributes (e.g., xsi:someName) were not being deserialized because the namespace-aware XML parser stores them by local name (someName), not the prefixed form.
- XML test comparator: Changed child node comparison to be order-independent, since XML element order.

### Testing

Used AmazonSSM to test codegen, the client builder adds the service schema as expected

```java
    /**
     * Builder for {@link AmazonSSMClient}.
     */
    final class Builder extends Client.Builder<AmazonSSMClient, Builder> {
        private static final ProtocolSettings protocolSettings = ProtocolSettings.builder()
                .service(ShapeId.from("com.amazonaws.ssm#AmazonSSM"))
                .serviceVersion("2014-11-06")
                .serviceSchema(AmazonSSMApiService.instance().schema())
                .build();
        private static final AwsJson1_1Trait protocolTrait = new AwsJson1_1Trait.Provider().createTrait(
            ShapeId.from("aws.protocols#awsJson1_1"),
            Node.objectNode()
        );
```

--

### DIFF
--- a/client/client-core/src/main/java/software/amazon/smithy/java/client/core/ProtocolSettings.java
+++ b/client/client-core/src/main/java/software/amazon/smithy/java/client/core/ProtocolSettings.java
@@ -84,6 +84,9 @@ public final class ProtocolSettings {
          * @return the builder
          */
         public Builder serviceSchema(Schema serviceSchema) {
+            if (service == null && serviceSchema != null) {
+                service = serviceSchema.id();
+            }
             this.serviceSchema = serviceSchema;
             return this;
         }
--- a/client/client-core/src/main/java/software/amazon/smithy/java/client/core/ProtocolSettings.java
+++ b/client/client-core/src/main/java/software/amazon/smithy/java/client/core/ProtocolSettings.java
@@ -5,6 +5,7 @@
 
 package software.amazon.smithy.java.client.core;
 
+import software.amazon.smithy.java.core.schema.Schema;
 import software.amazon.smithy.model.shapes.ShapeId;
 
 /**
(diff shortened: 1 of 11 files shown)


client/client-core/src/main/java/software/amazon/smithy/java/client/core/ProtocolSettings.java
```diff
@@ -61,6 +77,17 @@ public Builder serviceVersion(String serviceVersion) {
...
         }
 
+        /**
+         * Sets the service schema.
+         *
+         * @param serviceSchema the service schema
+         * @return the builder
+         */
+        public Builder serviceSchema(Schema serviceSchema) {
+            this.serviceSchema = serviceSchema;
```
mtdowling, line 90: set the service ID here if it's null?

mtdowling (approved):

PR #1167, opened 2026-05-02
Add a mechanism to avoid mixing versions for a client

sugmanue:
*Issue #, if available:*

*Description of changes:*

Other SDKs have historically suffered from issues where users inadvertently mix versions of runtime libraries in their final application. This typically happens when different dependencies pull in different versions of the same module, resulting in a classpath that contains, for example, core at version 1.1.0 alongside client-http at version 1.0.9. These mismatches cause subtle runtime errors, class not found exceptions, missing methods, or unexpected behavior, that are difficult to diagnose because nothing fails at compile time. 

This change adds a version compatibility test that addresses this at runtime when the codegen schema classes are loaded: if the modules on the classpath don't all agree on the same version, or if any module is older than what the client was generated against, the test fails with a clear message identifying exactly which modules are mismatched, which prevents the classloader from succeeding. 

The end-user can set the java property `smithy.java.skipVersionCheck` to bypass this check if needed.

---


codegen/codegen-core/src/main/resources/software/amazon/smithy/java/codegen/integrations/version/SmithyVersionCompatibilityTest.java.template
```diff
@@ -0,0 +1,87 @@
...
+                props.load(is);
+            }
+            modules.add(new String[]{
+                props.getProperty("module", "unknown"),
+                props.getProperty("version", "unknown")
+            });
+        }
+
+        if (modules.isEmpty()) {
+            fail("No Smithy Java version markers found on the classpath. "
```
mtdowling: Does this make setting up a package more involved? Like do you have to ensure you add:

```
compileTestJava {
    dependsOn(smithyBuild)
}
```

If that's the case, wondering if this should warn instead

codegen/codegen-core/src/main/resources/software/amazon/smithy/java/codegen/integrations/version/SmithyVersionCompatibilityTest.java.template
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Properties;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.fail;
+
+/**
+ * Auto-generated test that validates all Smithy Java modules on the classpath
+ * have compatible versions. Generated against version {@code ${VERSION}}.
+ */
+public final class SmithyVersionCompatibilityTest {
```
adwsingh: Why have a test instead of just failing startup? We could put this at the top of `Schemas.java` or in the Client/Server initialization.

adwsingh (changes requested):

codegen/codegen-core/src/main/resources/software/amazon/smithy/java/codegen/integrations/version/SmithyVersionCompatibilityTest.java.template
sugmanue: We could but I wanted to avoid impacting cold start for some use cases where this matter. Probably not a big deal.

sugmanue: Also, test fails at build time, failing earlier is also good.

codegen/codegen-core/src/main/resources/software/amazon/smithy/java/codegen/integrations/version/SmithyVersionCompatibilityTest.java.template
sugmanue: Yes, it does, and also easy to miss, but I think that's OK provided that we deliver on the planned gradle plugin that will take care of this and all the setup that's currently needed.

codegen/codegen-core/src/main/resources/software/amazon/smithy/java/codegen/integrations/version/SmithyVersionCompatibilityTest.java.template
sugmanue: Ok, done as requested, the test run nows when the codegen class is loaded.

codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/generators/SchemasGenerator.java
```diff
@@ -63,7 +68,10 @@ public void accept(CustomizeDirective<CodeGenerationContext, JavaCodegenSettings
                                          * Defines schemas for shapes in the model package.
                                          */
                                         final class ${className:L} {
-                                            ${#builders}${value:C|}
+                                            ${?versionCheck}static {
+                                                ${versionCheck:C|}
+                                            }
+                                            ${/versionCheck}${#builders}${value:C|}
```
adwsingh: I don’t love doing this during class initialization of code generated classes.

1. Failures during class init are painful to debug. The first access shows up as the dreaded ExceptionInInitializerError, and every access after that turns into a NoClassDefFoundError. At that point users have to go spelunking through older logs to find the original failure, which is usually nowhere near the code path they’re currently debugging.
2. This also does not work cleanly for cases like Dynamic Client.

My preference would be to add this to the ClientBuilder instead.

core/src/main/java/software/amazon/smithy/java/core/VersionCheck.java
```diff
@@ -0,0 +1,138 @@
...
+                    SKIP_PROPERTY);
+            return;
+        }
+
+        var modules = new ArrayList<String[]>();
+        try {
+            var urls = Thread.currentThread()
+                    .getContextClassLoader()
+                    .getResources(VERSIONS_RESOURCE);
+            while (urls.hasMoreElements()) {
```
adwsingh: I think we can do a much cheaper check before scanning the classpath by comparing this version to the core's Version.VERSION, since in most cases if core doesn't match with the codegen version others won't match either.

buildSrc/src/main/kotlin/GenerateVersionProviderTask.kt
```diff
@@ -0,0 +1,33 @@
...
+    var moduleVersion: String = ""
+
+    @get:OutputDirectory
+    var outputDir: File = project.layout.buildDirectory.dir("generated/version-provider").get().asFile
+
+    @TaskAction
+    fun generate() {
+        val dir = File(outputDir, "resources/META-INF/smithy-java")
+        dir.mkdirs()
+        File(dir, "versions.properties").writeText(
```
adwsingh: Food for thought: I think SPIs would be a better fit than a versions.properties file, especially for shaded/fat-jar cases.

With a fat jar, all the versions.properties files can get collapsed into a single file, and then you’re in “last writer wins” territory. That’s pretty fragile and also hard to notice when it breaks.

An SPI avoids that class of problem. Each dependency contributes its own service entry, and most common shading tools already know how to merge service files correctly, including Gradle Shadow and maven-shade-plugin. So instead of relying on resource overwrite behavior, we get a model that composes much more naturally under shading.

An SPI also allows us to do fancy stuff like declare methods like `lastCompatibleVersion` so that we can have more complex checks than version equality which would be a hedge against https://github.com/smithy-lang/smithy-java/pull/1167/changes#r3211780970

core/src/jmh/java/software/amazon/smithy/java/core/VersionCheckBench.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Setup
+    @SuppressFBWarnings(value = "LG_LOST_LOGGER_DUE_TO_WEAK_REFERENCE", justification = "Intentional for benchmark")
+    public void setup() {
+        Logger.getLogger(VersionCheck.class.getName()).setLevel(Level.OFF);
+    }
+
+    @Benchmark
+    public void versionCheckEnabled() {
+        VersionCheck.check(VERSION);
```
adwsingh: I don't think this is truly measuring what we want it to measure which is how much does this add to the cold start cost.

core/src/main/java/software/amazon/smithy/java/core/VersionCheck.java
```diff
@@ -0,0 +1,138 @@
...
+        if (modules.isEmpty()) {
+            return;
+        }
+
+        var errors = new ArrayList<String>();
+
+        // All modules must report the same version.
+        var firstVersion = modules.get(0)[1];
+        for (var module : modules) {
+            if (!module[1].equals(firstVersion)) {
```
adwsingh: I am a little hesitant of strict version equality specially because we follow semver and this would complicate customers deploying security patches. Like if there is a security issue in one of the modules, we force people to upgrade all of them. Not saying that might be a bad thing, just letting you know the ramifications of this decision.

adwsingh (changes requested):

buildSrc/src/main/kotlin/GenerateVersionProviderTask.kt
sugmanue: switched to use SPI

core/src/jmh/java/software/amazon/smithy/java/core/VersionCheckBench.java
sugmanue: elaborate?

core/src/main/java/software/amazon/smithy/java/core/VersionCheck.java
```diff
@@ -0,0 +1,122 @@
...
+            return;
+        }
+
+        var modules = new ArrayList<ModuleVersion>();
+        for (var provider : ServiceLoader.load(SmithyVersionProvider.class)) {
+            modules.add(provider.getModuleVersion());
+        }
+
+        if (modules.isEmpty()) {
+            return;
```
adwsingh, line 65: You should set `CHECKED` to true here too

adwsingh (approved):

Issue #1089, opened 2026-03-19
Allow custom validation rules in validator


timocov:
Currently the validator only validates against known traits (e.g. `range` etc) and if you defined a custom protocol or just added some custom traits, there is no way to add validation rules for them unless you implement the whole validation yourself. It would be great if `Validator` would provide a way to override handling for each node so that if there is a custom trait then it can be handled. I see it maybe like an extra class (maybe similar to `ShapeVisitor` with `ShapeVisitor.Default` option but with validation API) that you can provide to the validator that it will call for every node it sees before/after the default validation. I know that the generated schema extracts some of the "known" values and saves them in the schema for performance reasons I guess and maybe it would be good to have an option to provide custom extracted fields as well? But I'm not too sure how it going to look like tho. Let me know what you think, I'm happy to contribute (if we agree on the solution).

mtdowling:
Any solution here needs to ensure optimal validation performance for most users. An extension mechanism would need to be near zero
cost when not in use. If that isn’t possible, then another validation pass for custom constraints might be the solution for this (eg rather than an expensive visitor).

Gravewalker666:
Hey @mtdowling I've opened a PR (#1171) proposing a solution for this issue. Will appreciate if you could take a look :)

mtdowling:
Thanks @Gravewalker666! This is now merged.

Gravewalker666:
Thanks @mtdowling! Appreciate the prompt reviews and having it merged!

PR #1171, opened 2026-05-07
Introduce custom constraints to extend the smithy validation rules

Gravewalker666:
*Issue #, if available:*
#1089 

*Description of changes:*
- Introduced `CustomConstraint` interface for extensible validation rules via ServiceLoader                             
- Added `Validator.CustomConstraintProvider` to discover and register custom constraints as schema extensions         
- Integrated `applyCustomConstraints()` into Validator.ShapeValidator to apply custom rules after standard validation
- Added `ValidationError.CustomValidationFailure` record for custom constraint violations

*Benchmark numbers:*
```
  Benchmark                                      Baseline (ns/op)    Feature (ns/op)    Improvement
  StringValidationBench.stringComposite          53.19               51.27              +3.6%                            
  StringValidationBench.stringEnumOnly           11.14               11.18              -0.4%                          
  StringValidationBench.stringLengthOnly         10.28               10.92              -6.2%
  StringValidationBench.stringNoValidation        9.90                9.90               0.0%
  StringValidationBench.stringPatternOnly        51.39               52.44              -2.0%
  ValidatorBench.person                          15.74               15.76              -0.1%
  ValidatorBench.pojoWithValidatedCollections   150.44              146.93              +2.3%
  ValidatorBench.unvalidatedPojo                 15.05               12.91             +14.2%
  ValidatorBench.validatedPojo                   17.83               14.01             +21.4%

```

A record of benchmark runs I've done along the way is documented here (non-complete): https://pastebin.com/GTuJKkQN

### DIFF
--- a/core/src/main/java/software/amazon/smithy/java/core/schema/CustomConstraint.java
+++ b/core/src/main/java/software/amazon/smithy/java/core/schema/CustomConstraint.java
@@ -8,16 +8,16 @@ package software.amazon.smithy.java.core.schema;
 import java.util.List;
 
 /**
- * A custom validation rule that extends Smithy's validation framework.
+ * A custom constraint that extends Validation API.
  *
  * <p>Implementations are discovered via {@link java.util.ServiceLoader} and must be registered in
- * {@code META-INF/services/software.amazon.smithy.java.core.schema.CustomValidationRule}.
+ * {@code META-INF/services/software.amazon.smithy.java.core.schema.CustomConstraint}.
  *
  * @see Validator
- * @see Validator.CustomValidationRuleProvider
+ * @see Validator.CustomConstraintProvider
  * @see ValidationError.CustomValidationFailure
  */
-public interface CustomValidationRule {
+public interface CustomConstraint {
     /**
      * Determines whether this rule applies to the given schema.
      *
--- a/core/src/main/java/software/amazon/smithy/java/core/schema/CustomConstraint.java
+++ b/core/src/main/java/software/amazon/smithy/java/core/schema/CustomConstraint.java
@@ -23,8 +23,6 @@ public interface CustomConstraint {
     /**
      * Returns the set of shape types this constraint applies to.
      *
-     * <p>Return an empty set to match all shape types
-     *
      * @return the set of shape types this constraint validates, or an empty set for all types
      */
     EnumSet<ShapeType> appliesTo();
--- a/core/src/main/java/software/amazon/smithy/java/core/schema/CustomConstraint.java
+++ b/core/src/main/java/software/amazon/smithy/java/core/schema/CustomConstraint.java
@@ -17,13 +17,13 @@ import software.amazon.smithy.model.shapes.ShapeType;
  * {@code META-INF/services/software.amazon.smithy.java.core.schema.CustomConstraint}.
  *
  * @see Validator
- * @see ValidationError.CustomValidationFailure
+ * @see ValidationError
  */
 public interface CustomConstraint {
     /**
      * Returns the set of shape types this constraint applies to.
      *
-     * <p>Return an empty set to match all shape types (wildcard).
+     * <p>Return an empty set to match all shape types
      *
      * @return the set of shape types this constraint validates, or an empty set for all types
      */
--- a/core/src/main/java/software/amazon/smithy/java/core/schema/CustomConstraint.java
+++ b/core/src/main/java/software/amazon/smithy/java/core/schema/CustomConstraint.java
@@ -5,7 +5,10 @@
 
 package software.amazon.smithy.java.core.schema;
 
+import java.util.EnumSet;
 import java.util.List;
+import java.util.function.Supplier;
+import software.amazon.smithy.model.shapes.ShapeType;
 
 /**
  * A custom constraint that extends Validation API.
@@ -14,25 +17,25 @@ import java.util.List;
  * {@code META-INF/services/software.amazon.smithy.java.core.schema.CustomConstraint}.
  *
  * @see Validator
- * @see Validator.CustomConstraintProvider
  * @see ValidationError.CustomValidationFailure
  */
 public interface CustomConstraint {
     /**
-     * Determines whether this rule applies to the given schema.
+     * Returns the set of shape types this constraint applies to.
      *
-     * @param schema the schema to check
-     * @return {@code true} if this rule should validate values of this schema
+     * <p>Return an empty set to match all shape types (wildcard).
+     *
+     * @return the set of shape types this constraint validates, or an empty set for all types
      */
-    boolean appliesTo(Schema schema);
+    EnumSet<ShapeType> appliesTo();
 
     /**
      * Validates the given value against this custom rule.
      *
      * @param schema the schema of the value being validated
      * @param value the value to validate
-     * @param path the path to the value (e.g., "/user/address/zipCode")
+     * @param pathSupplier a supplier that lazily computes the path to the value (e.g., "/user/address/zipCode")
      * @return a list of validation errors, or an empty list if validation passes
      */
-    List<ValidationError> validate(Schema schema, Object value, String path);
+    List<ValidationError> validate(Schema schema, Object value, Supplier<String> pathSupplier);
 }
--- a/core/src/main/java/software/amazon/smithy/java/core/schema/ValidationError.java
+++ b/core/src/main/java/software/amazon/smithy/java/core/schema/ValidationError.java
@@ -137,4 +137,6 @@ public interface ValidationError {
             return "Conflicting list item found at position " + position;
         }
     }
+
+    record CustomValidationFailure(String path, Schema schema, String message) implements ValidationError { }
 }
--- a/core/src/main/java/software/amazon/smithy/java/core/schema/ValidationError.java
+++ b/core/src/main/java/software/amazon/smithy/java/core/schema/ValidationError.java
@@ -137,6 +137,4 @@ public interface ValidationError {
             return "Conflicting list item found at position " + position;
         }
     }
-
-    record CustomValidationFailure(String path, Schema schema, String message) implements ValidationError {}
 }
--- a/core/src/main/java/software/amazon/smithy/java/core/schema/ValidationError.java
+++ b/core/src/main/java/software/amazon/smithy/java/core/schema/ValidationError.java
@@ -138,5 +138,5 @@ public interface ValidationError {
         }
     }
 
-    record CustomValidationFailure(String path, Schema schema, String message) implements ValidationError { }
+    record CustomValidationFailure(String path, Schema schema, String message) implements ValidationError {}
 }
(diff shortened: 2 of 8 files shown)


core/src/main/java/software/amazon/smithy/java/core/schema/ValidationError.java
```diff
@@ -137,4 +137,6 @@ private static String createMessage(int position) {
             return "Conflicting list item found at position " + position;
         }
     }
+
+    record CustomValidationFailure(String path, Schema schema, String message) implements ValidationError {}
```
timocov: I'm not sure this needs to be defined here, `ValidationError` is an open public interface that custom integrations can implement and define their own errors in their packages? Feels like redundant to me

core/src/main/java/software/amazon/smithy/java/core/schema/Validator.java
```diff
@@ -488,5 +505,58 @@ private void checkType(Schema schema, ShapeType type) {
...
+                return;
+            }
+            var customConstraints = schema.getExtension(CustomConstraintProvider.CUSTOM_CONSTRAINT_EXTENSION_KEY);
+            if (customConstraints == null) {
+                return;
+            }
+            for (var rule: customConstraints) {
+                var validationErrors = rule.validate(schema, value, createPath());
+                for (var error: validationErrors) {
+                    addError(error);
```
timocov: It might worth adding a new `addErrors` method as adding multiple objects should be faster than adding then one by one. Also, having it this way might lead to the situation where, say, you have `maxAllowedErrors=1` but custom validator returned 2 errors (and these are the only errors during the validation) - we will add only first but the second one has been generated already so maybe we should still add it? It feels like `maxAllowedErrors` is more like a circuit breaker (to stop further validation) but if the validation is done already, why not just add these errors?

core/src/main/java/software/amazon/smithy/java/core/schema/Validator.java
```diff
@@ -488,5 +505,58 @@ private void checkType(Schema schema, ShapeType type) {
                 throw new ValidationShortCircuitException();
             }
         }
+
+        private void applyCustomConstraints(Schema schema, Object value) {
+            if (!CustomConstraintProvider.HAS_CUSTOM_CONSTRAINTS) {
```
timocov: this feels a bit odd to me - we have a static field but we operate on a instance basis. Should either this to be related to an instance of `CustomConstraintProvider` or other methods (e.g. provide, key) to be static in the class?

core/src/main/java/software/amazon/smithy/java/core/schema/Validator.java
```diff
@@ -488,5 +505,58 @@ private void checkType(Schema schema, ShapeType type) {
...
+                return;
+            }
+            var customConstraints = schema.getExtension(CustomConstraintProvider.CUSTOM_CONSTRAINT_EXTENSION_KEY);
+            if (customConstraints == null) {
+                return;
+            }
+            for (var rule: customConstraints) {
+                var validationErrors = rule.validate(schema, value, createPath());
```
mtdowling: createPath should be created up front and reused. Or better still, we could give the validator a Supplier instead of eagerly computing the path.

core/src/main/java/software/amazon/smithy/java/core/schema/Validator.java
```diff
@@ -260,6 +262,7 @@ public <T> void writeList(Schema schema, T state, int size, BiConsumer<T, ShapeS
 
                 checkListLength(schema, count);
             }
+            applyCustomConstraints(schema, state);
```
mtdowling, line 289: Technically, state passed to writeList and writeMap don't have to be the list/map being written (e.g., it could be streaming, or just canned content). The "state" value just acts as a carrier to avoid needing a closure that references up-values. That makes custom list and map validation tricky.

core/src/main/java/software/amazon/smithy/java/core/schema/Validator.java
```diff
@@ -488,5 +505,58 @@ private void checkType(Schema schema, ShapeType type) {
...
+            if (!CustomConstraintProvider.HAS_CUSTOM_CONSTRAINTS) {
+                return;
+            }
+            var customConstraints = schema.getExtension(CustomConstraintProvider.CUSTOM_CONSTRAINT_EXTENSION_KEY);
+            if (customConstraints == null) {
+                return;
+            }
+            for (var rule: customConstraints) {
+                var validationErrors = rule.validate(schema, value, createPath());
+                for (var error: validationErrors) {
```
mtdowling:
```suggestion
                for (var error : validationErrors) {
```

core/src/main/java/software/amazon/smithy/java/core/schema/Validator.java
```diff
@@ -488,5 +505,58 @@ private void checkType(Schema schema, ShapeType type) {
...
+        }
+
+        @Override
+        public SchemaExtensionKey<List<CustomConstraint>> key() {
+            return CUSTOM_CONSTRAINT_EXTENSION_KEY;
+        }
+
+        @Override
+        public List<CustomConstraint> provide(Schema schema) {
+            var rulesForThisSchema = new ArrayList<CustomConstraint>();
```
mtdowling: This seems heavy. It adds a list to every schema we validate. I think instead of this, we could have CustomConstraint return an `EnumSet<ShapeType> appliesTo()`. Then we'd have a `List<CustomConstraint>[]` on the class, sized to the number of shape types. Then check if there's an entry per/shape being validated. Then there's no SchemaExtension and no need to punch an array/schema onto all schemas.

mtdowling (changes requested):

core/src/main/java/software/amazon/smithy/java/core/schema/Validator.java
timocov: > Then we'd have a List<CustomConstraint>[] on the class, sized to the number of shape types

@mtdowling does it mean that there would be only one custom constraint possible per shape type? If so, which approach you'd choose to select which one should apply when there are multiple options? I'm curious if in this case the overall approach should be to pass "constraint per shape type" map to the validator constructor to avoid ambiguity when using SPI...

mtdowling: I think we should do away with the idea of per/shape custom validators, and instead have per/type custom validators (which there could be many per/type). Maybe with support for a wildcard which matches any type. If you need shape-specific validators (which would likely be rare), then that filtering can be done in `validate` by inspecting the Schema.

core/src/main/java/software/amazon/smithy/java/core/schema/Validator.java
Gravewalker666: Thanks for pointing that out. 

Now that we are applying the validations by the type there's a lot that falls on how the validation function itself is being written, you as the writer of a custom validator should select the instances of the particular type carefully before applying the validation logic. 

For this particular case it's checking the type of object they get, a list/map, or a stream, or canned content,  and write the validation accordingly. 

Am I write to assume that this is more of problem of how we can make them aware of this behaviour?

core/src/main/java/software/amazon/smithy/java/core/schema/Validator.java
```diff
@@ -120,6 +122,37 @@ static final class ValidationShortCircuitException extends SerializationExceptio
...
+                    for (var i = 0; i < CUSTOM_CONSTRAINTS_BY_TYPE.length; i++) {
+                        addConstraint(i, constraint);
+                    }
+                } else {
+                    for (var type : types) {
+                        addConstraint(type.ordinal(), constraint);
+                    }
+                }
+            }
+            HAS_CUSTOM_CONSTRAINTS = Arrays.stream(CUSTOM_CONSTRAINTS_BY_TYPE).anyMatch(java.util.Objects::nonNull);
```
mtdowling: nit - you can just track this with a boolean during the addConstraint loop above, then set HAS_CUSTOM_CONSTRAINTS to the boolean value that is set to true any time a custom constraint was found

core/src/main/java/software/amazon/smithy/java/core/schema/Validator.java
```diff
@@ -488,5 +537,22 @@ private void checkType(Schema schema, ShapeType type) {
...
+        private void applyCustomConstraints(Schema schema, Object value) {
+            if (!HAS_CUSTOM_CONSTRAINTS) {
+                return;
+            }
+
+            // Get constraints for this specific type
+            var typeConstraints = CUSTOM_CONSTRAINTS_BY_TYPE[schema.type().ordinal()];
+            if (typeConstraints != null) {
+                for (var constraint : typeConstraints) {
+                    var validationErrors = constraint.validate(schema, value, this::createPath);
```
mtdowling: I'm not 100% sure hotspot will hoist `this::createPath`, so it may be worth moving it to a local variable before the loop here.

core/src/test/java/software/amazon/smithy/java/core/schema/TestCustomConstraints.java
```diff
@@ -0,0 +1,116 @@
...
+
+public final class TestCustomConstraints {
+
+    private TestCustomConstraints() {}
+
+    public static class AlwaysFailsConstraint implements CustomConstraint {
+        @Override
+        public EnumSet<ShapeType> appliesTo() {
+            // Empty set means wildcard - applies to all types
+            return EnumSet.noneOf(ShapeType.class);
```
mtdowling: I think `EnumSet.allOf(ShapeType.class)` would be a better "wildcard", and would require no special casing wrt an empty set

mtdowling (changes requested):

core/src/main/java/software/amazon/smithy/java/core/schema/Validator.java
mtdowling: Yeah, it's probably fine to do it this way then

core/src/main/java/software/amazon/smithy/java/core/schema/CustomConstraint.java
```diff
@@ -0,0 +1,41 @@
...
+ * {@code META-INF/services/software.amazon.smithy.java.core.schema.CustomConstraint}.
+ *
+ * @see Validator
+ * @see ValidationError
+ */
+public interface CustomConstraint {
+    /**
+     * Returns the set of shape types this constraint applies to.
+     *
+     * <p>Return an empty set to match all shape types
```
mtdowling: This comment is stale now

mtdowling (changes requested):
Looks great! I'll merge after the stale comment fix.

core/src/main/java/software/amazon/smithy/java/core/schema/CustomConstraint.java
Gravewalker666: fixed now :)

mtdowling (approved):

PR #1176, opened 2026-05-14
Add event stream protocol tests for client

joewyz:
Add event stream protocol tests for client. Will work on fixes for skipped test cases later.

### DIFF
--- a/aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/AwsEventShapeDecoder.java
+++ b/aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/AwsEventShapeDecoder.java
@@ -170,7 +170,8 @@ final class AwsEventShapeDecoder<E extends SerializableStruct, IR extends Serial
         public <T> void readStruct(Schema schema, T builder, ShapeDeserializer.StructMemberConsumer<T> consumer) {
             var payloadWritten = false;
             for (Schema member : schema.members()) {
-                if (member.hasTrait(TraitKey.EVENT_HEADER_TRAIT)) {
+                if (member.hasTrait(TraitKey.EVENT_HEADER_TRAIT)
+                        && headersDeserializer.headers.containsKey(member.memberName())) {
                     consumer.accept(builder, member, headersDeserializer);
                 } else if (member.hasTrait(TraitKey.EVENT_PAYLOAD_TRAIT)) {
                     consumer.accept(builder, member, codecDeserializer);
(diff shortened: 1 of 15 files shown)


aws/client/aws-client-restjson/src/it/java/software/amazon/smithy/java/client/aws/restjson/RestJson1ProtocolTests.java
```diff
@@ -26,6 +27,40 @@
         skipOperations = {
                 // We dont ignore defaults on input shapes
                 "aws.protocoltests.restjson#OperationWithDefaults",
+        },
+        skipTests = {
```
mtdowling, line 31: Is this list right? We had no exclusions before and now we have a ton?

aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/AwsEventShapeDecoder.java
```diff
@@ -170,7 +170,8 @@ static class EventStreamDeserializer extends SpecificShapeDeserializer {
         public <T> void readStruct(Schema schema, T builder, ShapeDeserializer.StructMemberConsumer<T> consumer) {
             var payloadWritten = false;
             for (Schema member : schema.members()) {
-                if (member.hasTrait(TraitKey.EVENT_HEADER_TRAIT)) {
+                if (member.hasTrait(TraitKey.EVENT_HEADER_TRAIT)
+                        && headersDeserializer.headers.containsKey(member.memberName())) {
```
mtdowling, line 174: Why do we need this?

aws/client/aws-client-restjson/src/it/java/software/amazon/smithy/java/client/aws/restjson/RestJson1ProtocolTests.java
joewyz: Most of the skipped blob tests can be removed given that we released the [change](https://github.com/smithy-lang/smithy/pull/3121) to update the test model. Will work on the remaining fixes later.

aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/AwsEventShapeDecoder.java
joewyz: I added this to prevent it from reading all the members in the structure, otherwise all header input tests would fail.
example structure: 
```
structure HeadersEvent {
    @eventHeader
    booleanHeader: Boolean

    @eventHeader
    byteHeader: Byte

    @eventHeader
    shortHeader: Short

    @eventHeader
    intHeader: Integer

    @eventHeader
    longHeader: Long

    @eventHeader
    blobHeader: Blob

    @eventHeader
    stringHeader: String

    @eventHeader
    timestampHeader: DateTime
}
```

sugmanue (approved):

PR #1182, opened 2026-05-18
Optimize HTTP binding and XML overhead

mtdowling:
*Issue #, if available:*

*Description of changes:*


http/http-binding/src/main/java/software/amazon/smithy/java/http/binding/HttpBindingSerializer.java
```diff
@@ -48,104 +48,133 @@ final class HttpBindingSerializer extends SpecificShapeSerializer implements Sha
...
             shapeBodySerializer = payloadCodec.createSerializer(shapeBodyOutput);
-            // Serialize only the body members to the codec.
-            ShapeUtils.withFilteredMembers(schema, struct, this::bodyBindingPredicate)
-                    .serialize(shapeBodySerializer);
-            headers.put("content-type", List.of(payloadMediaType));
+            // Serialize only the body members to the codec, going through writeStruct so the codec emits the
+            // struct envelope (e.g. JSON {...}). The BodyMemberSerializer routes each member to the codec's
+            // per-member serializer (passed inside writeStruct) when its binding is BODY, and to a null serializer
+            // otherwise so non-body members (headers, labels, etc.) don't leak into the body.
+            shapeBodySerializer.writeStruct(schema, new StructBodyProxy(struct, activeBindings));
```
adwsingh: You can use `payloadCodec.serialize` here to take advantage of the codec's pooled serialize path. This should also let you get rid of storing both `shapeBodySerializer` and `shapeBodyOutput` on the class and just store a `ByteBuffer shapeBodyBuffer`

http/http-binding/src/main/java/software/amazon/smithy/java/http/binding/HttpBindingSchemaExtensions.java
```diff
@@ -0,0 +1,707 @@
...
+import software.amazon.smithy.utils.SmithyInternalApi;
+
+/**
+ * Pre-computes HTTP binding data on Schema objects at construction time so the per-request hot path can avoid
+ * trait lookups, type checks, and dispatch on a generic switch.
+ *
+ * <p>All internals are package-private and can be changed freely.
+ */
+@SmithyInternalApi
+public final class HttpBindingSchemaExtensions
```
adwsingh, line 35: You can also store `hasStructPayload` in this and avoid iterating through all the schema members [here](https://github.com/smithy-lang/smithy-java/blob/main/aws/client/aws-client-restjson/src/main/java/software/amazon/smithy/java/aws/client/restjson/RestJsonClientProtocol.java#L121-L128)

Another idea is you can also extend this to if a structure has no body then skip the call the codec once and cache the results and skip the codec entirely on subsequent calls.

adwsingh (approved):

PR #1183, opened 2026-05-18
Optimize CBOR Codec

adwsingh:
*Issue #, if available:*

*Description of changes:*

  - Replace Sink interface with direct byte[] buffer writes + VarHandle for multi-byte big-endian ops — no virtual dispatch per byte
  - Pool serializer instances via striped AtomicReferenceArray to avoid repeated buffer allocation
  - Pre-compute CBOR-encoded field name bytes at schema load time; write them with a single System.arraycopy
  - Inline UTF-8 encoding with an ASCII fast path: deprecated String.getBytes(int,int,byte[],int) bulk-copies, OR-accumulator detects non-ASCII, rewind path encodes UTF-8 directly into the buffer without allocation
  - Use a long bitmask instead of boolean[] for tracking indefinite-length collection nesting
  - Fuse ensureCapacity for blob writes — single capacity check covers CBOR header + payload, eliminating redundant buffer growth
  - Add Codec.serialize() that returns a ByteBuffer directly from the pooled serializer
 
Screenshot 2026-05-18 at 1 37 26 PM


codecs/cbor-codec/src/main/java/software/amazon/smithy/java/cbor/CborSchemaExtensions.java
```diff
@@ -0,0 +1,83 @@
...
+import software.amazon.smithy.model.shapes.ShapeType;
+
+/**
+ * Pre-computes CBOR codec data on Schema objects.
+ *
+ * <p>For member schemas: pre-encoded CBOR text string header + member name bytes.
+ * For struct/union schemas: {@link CborMemberLookup} instances for hash-based field matching
+ * and field name tables for O(1) lookup by memberIndex during serialization.
+ */
+public final class CborSchemaExtensions
```
mtdowling, line 23: Maybe mark `@SmithyInternalApi`

mtdowling (changes requested):

codecs/cbor-codec/src/main/java/software/amazon/smithy/java/cbor/CborMemberLookup.java
```diff
@@ -0,0 +1,88 @@
...
+import software.amazon.smithy.java.core.schema.Schema;
+
+/**
+ * Hash-based field name lookup that operates directly on UTF-8 bytes during CBOR deserialization.
+ *
+ * <p>Strategy:
+ * <ol>
+ *   <li><b>Speculative fast path</b>: check the expected next member via length + Arrays.equals
+ *       (JVM-intrinsified). Fires on every field when CBOR map entries arrive in schema definition
+ *       order (the common case for smithy-to-smithy communication).
```
rhernandez35, line 20: I avoided doing this originally because I didn't want adding new fields to have visible performance effects, especially to legacy clients. Adding a new optional field or reordering schema fields should not have a visible performance effect.

codecs/cbor-codec/src/main/java/software/amazon/smithy/java/cbor/CborMemberLookup.java
```diff
@@ -0,0 +1,88 @@
...
+            return orderedSchemas[expectedNext];
+        }
+
+        long hash = fnvHash(buf, start, end);
+        for (int i = 0; i < orderedHashes.length; i++) {
+            if (orderedHashes[i] == hash
+                    && orderedNameBytes[i].length == nameLen
+                    && Arrays.equals(buf, start, end, orderedNameBytes[i], 0, nameLen)) {
+                return orderedSchemas[i];
+            }
```
rhernandez35, line 74: This introduces a linear scan over all schema-defined keys for every unknown field. Even with the length checking, I would be wary of doing this for untrusted input.

rhernandez35 (dismissed):

rhernandez35:
Please do some perf testing with clients that have out-of-date schemas and servers that are receiving unknown fields. I'm concerned about adverse impact outside of all but the happiest cases in this PR.

codecs/cbor-codec/src/main/java/software/amazon/smithy/java/cbor/CborMemberLookup.java
adwsingh: I considered this and had local benchmarks for testing a fully reversed payload with unknown fields. I have now added it to the serde benchmarks too in https://github.com/smithy-lang/smithy-java/pull/1185.

Results on m7i.xlarge:

~16% at p50 for a fully reversed payload with unknown fields interspersed
  
 | Test Case                  | p50 (ns) | p99 (ns) |
  |----------------------------|----------|----------|
  | GetItemOutput_M   | 1710     | 2572     |
  | GetItemOutput_OutOfOrder   | 1984     | 2864     |


The scan stays pretty cheap because the 64-bit FNV hash is fast to compute and acts as a pre-filter, so for normal non-matches we bail out before hitting `Arrays.equals`. Only hash matches need the actual byte compare. Also, `N` here is the schema member count, usually single digits to low tens, not the number of fields in the incoming payload. The scan itself is over a contiguous `long[]`, so it is fairly cache-friendly and not pointer-chasing through random objects.

For untrusted input, I think the practical DoS surface is still payload size / serde work, not the per-field lookup cost where `N` is fixed by the schema. Unless the schema itself is huge, this lookup path doesn't feel like the scary part.

Full score here, (`GetItemOutput_M` is used as the base for `GetItemOutput_OutOfOrder`

Screenshot 2026-05-18 at 4 49 20 PM

mtdowling (approved):

PR #1200, opened 2026-05-21
Added client call decorator feature

### LINKED ISSUE #1197: Overriding `Client.call` method
_(Originally asked in https://github.com/smithy-lang/smithy-java/pull/1162#discussion_r3196051732)_

I'd want to ask the team's vision on [`Client.call` method](https://github.com/smithy-lang/smithy-java/blob/146172be114bcbbac51afdeee312b9a0decb4e5f/client/client-core/src/main/java/software/amazon/smithy/java/client/core/Client.java#L76). Currently it is marked as `protected` and I understand that the main purpose of this is to allow derived "impl" classes (generated by the codegen) to call it, but also in theory it is possible to override it during the codegen to have some extra logic in it (and still call the original method from the base-class). For example, lets say you're integrating an XRay-like library into clients so that every client call creates a sub-segment, then the implementation of the overridden method can look something like this:

```java
    <I extends SerializableStruct, O extends SerializableStruct> O call(
        I input,
        ApiOperation<I, O> operation,
        RequestOverrideConfig overrideConfig
    ) {
        // note that "ServiceName.OperationName" can be extracted from operation/service, but I omitted it from this example
        return AWSXRay.getGlobalRecorder().createSegment("ServiceName.OperationName", () -> {
            return super.call(input, operation, overrideConfig);
        });
    }
```

In this case, we're using library's provided "wrapping" function to make a call so that it can handle "start" and "end" correctly. In case of XRay, it provides `beginSegment`/`endSegment` methods so it is not quite fair comparison, but not all libraries have that and if it is a 3rd-party library it might not be even possible.

In my use-case, I have a library that manages current "context" state that then later is being used (or can be used) when making actual http requests at the transport level by adding some headers to request. And unfortunately this library has a couple of features so that even if I introduce begin-end-like methods, it won't be possible as it does some extra logic during this "call"-phase that uses library's internals.

So I have the following questions:
- Do you expect that consumers can/should be able to override `Client.call` method in their derived classes?
- If the answer is yes, would you accept a contribution of:
	- opening some of the `RequestOverrideConfig` fields to external consumers (or at least the context getter)? Currently every getter in this class is package-private so it is not possible to access them from the outside (e.g. to get the context which was previously set by custom settings set by `defaultSettings` (similar to https://github.com/smithy-lang/smithy-java/blob/b9d655953eeb5f886216ecf4657aac15f7143de8/examples/dynamodb-client/smithy-build.json#L9-L11). I understand that the main approach for this type of tasks should be using an interceptor, but when integrating into an existing system it is not always possible and some libraries require to wrap a call so that the context is correctly "scoped".
	- adding something (not yet sure what exactly, a codesection of some sort?) to the codegen so that it would be possible to do without parsing generated client class and finding a place to insert an override method in there? It is not critical thing I'd say as consumers still would be able to "hack it" from their side, but it would be a nice addition to reduce hacks in consumers code.

Thanks!

timocov:
*Issue #, if available:* fixes #1197

*Description of changes:*

Added support for client call decorator override described in https://github.com/smithy-lang/smithy-java/issues/1197#issuecomment-4504310850.

~Also marked `RequestOverrideConfig.context` so that it would become available in call decorators (happy to remove/change it if you think this is not what should be available).~

### DIFF
--- a/client/client-core/src/main/java/software/amazon/smithy/java/client/core/Client.java
+++ b/client/client-core/src/main/java/software/amazon/smithy/java/client/core/Client.java
@@ -82,7 +82,13 @@ public abstract class Client implements Closeable {
             RequestOverrideConfig overrideConfig
     ) {
         if (callDecorator != null) {
-            return callDecorator.apply(this, operation, input, overrideConfig, this::doCall);
+            return callDecorator.apply(
+                    this,
+                    operation,
+                    input,
+                    overrideConfig,
+                    overrideConfig == null ? null : overrideConfig.context(),
+                    this::doCall);
         }
 
         return this.doCall(input, operation, overrideConfig);
--- a/client/client-core/src/main/java/software/amazon/smithy/java/client/core/Client.java
+++ b/client/client-core/src/main/java/software/amazon/smithy/java/client/core/Client.java
@@ -43,7 +43,9 @@ public abstract class Client implements Closeable {
     private final ClientInterceptor interceptor;
     private final IdentityResolvers identityResolvers;
     private final RetryStrategy retryStrategy;
+    private final ClientCallDecorator<Client> callDecorator;
 
+    @SuppressWarnings("unchecked")
     protected Client(Builder<?, ?> builder) {
         ClientConfig.Builder configBuilder = builder.configBuilder();
         this.config = configBuilder.build();
@@ -61,6 +63,7 @@ public abstract class Client implements Closeable {
         if (retryStrategy instanceof Claimable c) {
             c.claim(this);
         }
+        this.callDecorator = (ClientCallDecorator<Client>) this.config.callDecorator();
     }
 
     /**
@@ -74,6 +77,18 @@ public abstract class Client implements Closeable {
      * @return Returns the deserialized output.
      */
     protected <I extends SerializableStruct, O extends SerializableStruct> O call(
+            I input,
+            ApiOperation<I, O> operation,
+            RequestOverrideConfig overrideConfig
+    ) {
+        if (callDecorator != null) {
+            return callDecorator.apply(this, operation, input, overrideConfig, this::doCall);
+        }
+
+        return this.doCall(input, operation, overrideConfig);
+    }
+
+    private <I extends SerializableStruct, O extends SerializableStruct> O doCall(
             I input,
             ApiOperation<I, O> operation,
             RequestOverrideConfig overrideConfig
@@ -367,7 +382,7 @@ public abstract class Client implements Closeable {
          * plugin class. Plugins are applied in a sorted {@link ClientPlugin.Phase} and insertion order.
          *
          * @see ClientConfig.Builder#pluginPredicate()
-         * @see ClientConfig.Builder#addPlugin(ClientPlugin) 
+         * @see ClientConfig.Builder#addPlugin(ClientPlugin)
          * @param plugin Plugin to add.
          * @return the builder.
          */
@@ -415,6 +430,18 @@ public abstract class Client implements Closeable {
                     return AutoPlugin.INSTANCE != plugin;
                 }
             }.and(configBuilder.pluginPredicate()));
+        }
+
+        /**
+         * Sets the decorator to wrap client call execution.
+         *
+         * @param callDecorator the client call decorator.
+         * @return the builder.
+         */
+        @SuppressWarnings("unchecked")
+        public B callDecorator(ClientCallDecorator<I> callDecorator) {
+            configBuilder.callDecorator(callDecorator);
+            return (B) this;
         }
 
         /**
--- a/client/client-core/src/main/java/software/amazon/smithy/java/client/core/RequestOverrideConfig.java
+++ b/client/client-core/src/main/java/software/amazon/smithy/java/client/core/RequestOverrideConfig.java
@@ -74,12 +74,7 @@ public final class RequestOverrideConfig {
         return identityResolvers;
     }
 
-    /**
-     * Get the context of the override request config.
-     *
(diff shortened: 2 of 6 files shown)


client/client-core/src/main/java/software/amazon/smithy/java/client/core/RequestOverrideConfig.java
```diff
@@ -74,7 +74,12 @@ List<IdentityResolver<?>> identityResolvers() {
         return identityResolvers;
     }
 
-    Context context() {
+    /**
+     * Get the context of the override request config.
+     *
+     * @return the context.
+     */
+    public Context context() {
```
timocov: One potential "issue" of this change I see is that call decorators now would be able to modify the context (so it is not a readonly), but maybe it is not an issue as the context is being exposed to hooks already where they could do the same before).

client/client-core/src/main/java/software/amazon/smithy/java/client/core/Client.java
```diff
@@ -77,6 +80,18 @@ protected <I extends SerializableStruct, O extends SerializableStruct> O call(
             I input,
             ApiOperation<I, O> operation,
             RequestOverrideConfig overrideConfig
+    ) {
+        if (callDecorator != null) {
+            return callDecorator.apply(this, operation, input, overrideConfig, this::doCall);
+        }
+
+        return this.doCall(input, operation, overrideConfig);
```
timocov, line 94: In theory this might be changed to have something like this in the constructor to avoid checking this field every time, let me know if you think this is better and I will update

```java
// in Client
private final ClientCallInvoker callInvoker;
```

```java
// in Client constructor
        var callDecorator = (ClientCallDecorator<Client>) this.config.callDecorator();
        if (callDecorator != null) {
            this.callInvoker = wrapDecorator(callDecorator);
        } else {
            this.callInvoker = this::doCall;
        }
```

```java
// in Client
    private ClientCallInvoker wrapDecorator(ClientCallDecorator<Client> decorator) {
        return new ClientCallInvoker() {
            @Override
            public <I extends SerializableStruct, O extends SerializableStruct> O invoke(I input,
                ApiOperation<I, O> operation, RequestOverrideConfig overrideConfig) {
                return decorator.apply(Client.this, operation, input, overrideConfig, Client.this::doCall);
            }
        };
    }
```

Then the `call` would be unconditional

```java
    protected <I extends SerializableStruct, O extends SerializableStruct> O call(
            I input,
            ApiOperation<I, O> operation,
            RequestOverrideConfig overrideConfig
    ) {
        return this.callInvoker.invoke(input, operation, overrideConfig);
    }
```

mtdowling: The if statement here is probably fine. We could always change it later if profiling shows it matters.

client/client-core/src/main/java/software/amazon/smithy/java/client/core/RequestOverrideConfig.java
mtdowling: Yeah, it currently kind of blurs the line between override and given context. Maybe we just add one more argument to the two interfaces and pass Context? It’s a lot of arguments, but seems necessary (and I want to avoid a kind of wrapper allocation for this since it’ll be a hot path).

mtdowling (changes requested):

client/client-core/src/main/java/software/amazon/smithy/java/client/core/RequestOverrideConfig.java
timocov: I was thinking about it and it felt like it might be a bit ambiguous as in hooks we have context that is "merged" (or whichever takes precedence) with client's default context, but here we're mostly interested in the override context? Maybe its just me.

Another thought, do you think we should keep the RequestOverrideConfig in the arguments of `ClientCallDecorator` and `ClientCallInvoker` despite the fact that you can't do anything with it outside of smithy-java other than pass it down? But on the other side removing it from there would require to create unnecessary wrapping around `doCall` method...

timocov: > Maybe its just me.

I guess thats fine since the implementation can always take client's current context and do its own logic for "overrides" if needed, at least for now.

timocov: @mtdowling pushed the changes in 1d6e17f and 669f55d

mtdowling: Thanks. Gonna take a deeper look at this. Wondering if we can just _not_ send RequestOverride somehow by resolving stuff before the decorator. Alternatively, expose ClientCall or a stripped down version of as an interface that exposes less than the internal ClientCall. Not sure.

mtdowling:
What do you think about this tweak? https://github.com/smithy-lang/smithy-java/pull/1211

timocov:
@mtdowling thanks for the update! Yes, your changes look good to me. I'm closing this PR in favor of yours 👍

The only "issue" I see is that now you can't 1) modify inputs (which is probably for good) 2) see what users provided in call overrides to differentiate between "client defaults" and "call overrides", but I'll probably cut another issue for this as it might be just completely separate thing (e.g. with `defaultSettings` we can provide settings interface that will be applied to both "client constructor settings" and "request overrides settings", which in case of default plugins constructions might be treated differently and you might want to separate the two - but I need to think about it more before cutting a ticket 😂)

mtdowling:
Good idea. I added a way to override input now using withInput.

PR #1203, opened 2026-05-22
Fix NPE in JsonCodec.deserializeShape that suppresses deserialization errors

### LINKED ISSUE #1202: JsonCodec.deserializeShape() NPE suppresses validation exceptions on invalid input
## Description

When using `JsonCodec.deserializeShape()` with input that contains a nested struct with missing required fields (e.g., inside a list), the original `SerializationException` ("Missing required members: [field]") is suppressed by an NPE from the `close()` method. Callers only see:

```
SerializationException: Cannot invoke "...JsonParser.currentToken()" because "this.parser" is null
```

This is a regression from 1.1.0, where the validation exception propagated correctly.

## Reproduction

```java
// Given a ComplexStruct with a `structList` field (list of NestedStruct),
// where NestedStruct has @required field1 (String) and field2 (Integer):
byte[] json = "{\"id\":\"x\",\"count\":1,\"enabled\":true,\"ratio\":1.0,\"score\":1.0,\"bigCount\":1,\"structList\":[{\"field1\":\"hello\"}]}"
        .getBytes(StandardCharsets.UTF_8);

var codec = JsonCodec.builder().build();
// Expected: SerializationException with message containing "field2"
// Actual: SerializationException wrapping NPE from close()
codec.deserializeShape(json, ComplexStruct.builder());
```

Root Cause
PR #1148 added `deserializer.close()` in a `finally` block in J`sonCodec.deserializeShape()` for Jackson buffer recycling. When deserialization fails mid-parse, `JacksonJsonDeserializer.close()`:

Calls `parser.nextToken()` — returns non-null (unconsumed tokens remain)
Calls `parser.close()` then sets `parser = null`
Calls `describeToken()` which does `parser.currentToken()` → NPE
The NPE from `close()` suppresses the original helpful exception.

Impact
Any caller that relies on the exception message from `deserializeShape()` to produce user-facing validation errors (e.g., HTTP 400 responses with field-specific messages) gets an unhelpful NPE instead.

Workaround
Use `createDeserializer()` directly to avoid the buggy close() path:

```
var deserializer = codec.createDeserializer(body);
var result = builder.deserialize(deserializer).errorCorrection().build();
// Do not call deserializer.close()
```

etspaceman:
*Issue #, if available:*

Fixes #1202

*Description of changes:*

When `deserializeShape()` encounters invalid input (e.g., a nested struct in a list with missing required fields), the builder throws a `SerializationException` mid-parse. The `finally` block in `JsonCodec.deserializeShape()` then calls `close()` on the deserializer, which triggers an NPE that suppresses the original helpful exception.

The NPE occurs in `JacksonJsonDeserializer.close()` because:
1. `parser.nextToken()` returns non-null (unconsumed tokens remain after the aborted parse)
2. `parser.close()` is called, then `parser = null`
3. `describeToken()` is called, which dereferences the now-null `parser` → NPE

**Fix:**
- In `JacksonJsonDeserializer.close()`, capture the token description before nulling the parser reference.
- In `JsonCodec.deserializeShape()`, use `closeQuietly()` on the error path so exceptions from `close()` don't suppress the original deserialization/validation exception.

**Test:** Added `deserializeShapeWithNestedMissingRequiredFieldReportsFieldName` which fails without this fix (NPE suppresses the "Missing required members" exception) and passes with it.

### DIFF
--- a/codecs/json-codec/src/main/java/software/amazon/smithy/java/json/JsonCodec.java
+++ b/codecs/json-codec/src/main/java/software/amazon/smithy/java/json/JsonCodec.java
@@ -56,20 +56,36 @@ public final class JsonCodec implements Codec {
     @Override
     public <T extends SerializableShape> T deserializeShape(byte[] source, ShapeBuilder<T> builder) {
         var deserializer = createDeserializer(source);
+        T result;
         try {
-            return builder.deserialize(deserializer).errorCorrection().build();
-        } finally {
-            deserializer.close();
+            result = builder.deserialize(deserializer).errorCorrection().build();
+        } catch (Exception e) {
+            closeQuietly(deserializer);
+            throw e;
         }
+        deserializer.close();
+        return result;
     }
 
     @Override
     public <T extends SerializableShape> T deserializeShape(ByteBuffer source, ShapeBuilder<T> builder) {
         var deserializer = createDeserializer(source);
+        T result;
         try {
-            return builder.deserialize(deserializer).errorCorrection().build();
-        } finally {
+            result = builder.deserialize(deserializer).errorCorrection().build();
+        } catch (Exception e) {
+            closeQuietly(deserializer);
+            throw e;
+        }
+        deserializer.close();
+        return result;
+    }
+
+    private static void closeQuietly(ShapeDeserializer deserializer) {
+        try {
             deserializer.close();
+        } catch (Exception ignored) {
+            // Don't suppress the original deserialization exception.
         }
     }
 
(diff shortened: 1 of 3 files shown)


etspaceman:
Whoops - forgot spotless. Updated.

adwsingh:
Thanks for the fix @etspaceman . Did you get a chance to take the new SmithyJson serde for a spin? Its opt in right now and we are looking for some real world feedback. Its much faster than Jackson see #1148 

You can opt in by setting the system property `-Dsmithy-java.json-provider=smithy`.

adwsingh (approved):

etspaceman:
No I haven't! Id love to give it a try. Will report back when I do.

PR #1206, opened 2026-05-22
Improve Java HTTP client integration

mtdowling:
Use buffers as-is without copying, add support for direct buffers to Jackson codec, improve HTTP header wrapper to use underlying data more, improve how we send data to Java client by adding a small adapter to avoid double-wrapping the publisher

*Issue #, if available:*

*Description of changes:*


io/src/main/java/software/amazon/smithy/java/io/datastream/MultiByteBufferDataStream.java
```diff
@@ -0,0 +1,168 @@
...
+    @Override
+    public boolean hasByteBuffer() {
+        // The contiguous buffer is materialized lazily by asByteBuffer().
+        return false;
+    }
+
+    @Override
+    public ByteBuffer asByteBuffer() {
+        var local = stitched;
+        if (local != null) {
```
adwsingh, line 87: While in general racy double-computations are fine, in this case it will lead to allocating a large byte buffer if the content-length is big.

client/client-http/src/main/java/software/amazon/smithy/java/client/http/JavaHttpClientTransport.java
```diff
@@ -71,28 +68,6 @@ private static void setHostProperties() {
...
-        }
-    }
-
-    /**
-     * Set JVM-global JDK HttpClient tuning defaults when the application has not already set
-     * them.
-     */
-    private static void setDefaultTuningProperties() {
-        setIfUnset("jdk.httpclient.maxframesize", "65536");
-        setIfUnset("jdk.httpclient.maxstreams", "1024");
```
adwsingh, line 94: Why did we remove this and switch to the default of 100?

adwsingh (approved):

PR #1229, opened 2026-06-05
Added support for getter sections in unions codegen

timocov:
### What behavior changes?

This PR adds `GetterSection` to getter of union variant classes.

One of the places where `GetterSection` helps is when you want to add custom annotations to getters. For instance, when integrating generated code with jakarta validation in Spring Boot, complex fields require to have `@Valid` annotation attached in order to be validated (e.g. to run full depth validation). Currently, `GetterSection` is not "emitted" which means that union member getters aren't being processed and it is possible to add `@Valid` annotation in this case.

### How was this validated?

I changed to code locally, run `gw pTML`, run the codegen, spin up a spring boot application and validated that the request is being validated correctly.

Should I use

```java
writer.pushState();
writer.injectSection(new GetterSection(member));
```

instead to avoid people accidentally removing the content of the method?

Also, should I do the same for enums?

### Additional Links

N/A.

---

### DIFF
--- a/codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/generators/UnionGenerator.java
+++ b/codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/generators/UnionGenerator.java
@@ -17,6 +17,7 @@ import software.amazon.smithy.java.codegen.CodegenUtils;
 import software.amazon.smithy.java.codegen.JavaCodegenSettings;
 import software.amazon.smithy.java.codegen.SymbolProperties;
 import software.amazon.smithy.java.codegen.sections.ClassSection;
+import software.amazon.smithy.java.codegen.sections.GetterSection;
 import software.amazon.smithy.java.codegen.writer.JavaWriter;
 import software.amazon.smithy.java.core.schema.Schema;
 import software.amazon.smithy.java.core.schema.SchemaUtils;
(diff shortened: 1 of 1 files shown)


adwsingh (approved):

adwsingh:
I wonder if the better design here is to generically be able to add support for custom Annotations in the codegen.

timocov:
@adwsingh do you mean something like adding a new property to symbols or something else? I think generally having this as a built-in feature would make sense as I think this is what it is being used for already (e.g. `@SmithyGenerated` is being added like that via `ClassSection` prepender). But I'm not sure how configurable this would be - in some cases you want to add annotations to classes, sometimes to methods (sometimes on fields, but I guess in most cases it can be replaced with methods instead).

mtdowling:
Yeah, a new property to symbols for this seems like it would be a better and more constrained DX. And maybe the annotations would have scoping rules when attached to a Symbol (getter, setter [builder], class, etc).

timocov:
I see. This might require some extra thinking on API I guess - initially I thought that just adding something like `Property<Symbol[]>` per type of annotation place (getter, setter, etc) would work, but then I realized that you also can provide arguments to annotations, that could be symbols on their own (e.g. you can reference a class as an argument to annotation), so in order to support annotations we might actually use writer to actually generate one (so that it will add imports etc).

I'm happy to work on it if we would have some directions/design for it, but if we don't, do you think it would be possible to get this one (or some variation of it) merged since it feels like it kinda fits into the set of features currently existing? (so it is more like a bug fix to some extent 🙃)

PR #1234, opened 2026-06-10
Add dynamic-client end-to-end test against real Connect model

yasmewad:
Loads the production AWS Connect model (DescribeContact, GET /contacts/{InstanceId}/{ContactId}) and exercises the full DynamicClient -> RestJsonClientProtocol -> PathSerializer path via a capturing transport (no credentials or network), asserting the two @httpLabel path variables serialize correctly.

This is an end-to-end smoke test for the dynamic-client path. Note: it does NOT reproduce the original ClassCastException on its own, because DocumentUtils.getMemberValue (since 4ec73e94) unwraps a string label via asObject() before PathSerializer sees it. The unit-level PathSerializerTest remains the regression guard for the cast bug; this test guards the dynamic-client integration end to end.

Adds smithy-aws-endpoints as a test dependency so the real Connect model (endpointBdd / aws.partition) assembles.

### What behavior changes?

### How was this validated?

### Additional Links

---


mtdowling:
I don't think we should ever merge this, so closing

PR #1235, opened 2026-06-10
Add a Gradle Plugin for codegen

adwsingh:
### What behavior changes?

Adds a Gradle plugin (`software.amazon.smithy.java.gradle.smithy-java`) that replaces manual `smithy-base` + `afterEvaluate` boilerplate with a single plugin application that auto-wires source sets, dependencies, and task ordering.

Every customer project currently duplicates ~30 lines of Gradle plumbing to wire generated code. This plugin makes the zero-config path a one-liner and provides an explicit DSL (`modes`, `generatedPluginOutputs`) for customization.

### How was this validated?

`./gradlew :mcp:mcp-schemas:sourcesJar :mcp:mcp-schemas:jar --rerun-tasks --configuration-cache` passes with correct service files in all artifacts. Configuration cache reuse confirmed on subsequent runs.

`SmithyJavaPlugin.java` (dependency wiring via `ValueSource`, service-file merge with `eachFile` source-path filtering) and `SmithyBuildModesValueSource.java` (configuration cache tracked file read).

### Additional Links

N/A

---


gradle-plugin/build.gradle.kts
```diff
@@ -0,0 +1,104 @@
...
+    mavenLocal()
+    mavenCentral()
+    gradlePluginPortal()
+}
+
+// Versions are pinned here because this module is an included build (standalone settings.gradle.kts)
+// and cannot access the parent project's version catalog. Update these when bumping smithy versions.
+dependencies {
+    implementation("software.amazon.smithy.gradle:smithy-base:1.4.0")
+    implementation("software.amazon.smithy:smithy-model:1.68.0")
```
timocov: should this be aligned with smithy-java's current version (which [is 1.71](https://github.com/smithy-lang/smithy-java/blob/dd53b24d1fbed7c1000c72d6693e720b76d22dd7/gradle/libs.versions.toml#L4))? Or maybe be open-ended like `[1.68, 2.0)`?

gradle-plugin/src/main/java/software/amazon/smithy/java/gradle/SmithyJavaPlugin.java
```diff
@@ -0,0 +1,255 @@
...
+public class SmithyJavaPlugin implements Plugin<Project> {
+
+    private static final String SMITHY_JAVA_GROUP = "software.amazon.smithy.java";
+    private static final String JAVA_CODEGEN_PLUGIN_NAME = "java-codegen";
+    private static final String SMITHY_BUILD_TASK_NAME = "smithyBuild";
+    private static final String MERGE_SERVICE_FILES_TASK_NAME = "mergeSmithyServiceFiles";
+
+    @Override
+    public void apply(Project project) {
+        project.getPlugins().apply(JavaLibraryPlugin.class);
```
timocov: In case of a server I believe this should be just `java` right?

gradle-plugin/src/main/java/software/amazon/smithy/java/gradle/SmithyJavaPlugin.java
```diff
@@ -0,0 +1,255 @@
...
+ * {@code software.amazon.smithy.gradle.smithy-base}, then automatically:
+ * <ul>
+ *     <li>Parses {@code smithy-build.json} to determine codegen modes</li>
+ *     <li>Adds required dependencies based on detected modes</li>
+ *     <li>Wires generated source and resource directories into the main source set</li>
+ *     <li>Sets up task dependencies (compileJava, processResources, sourcesJar)</li>
+ *     <li>Optionally merges META-INF/services files from multiple plugin outputs</li>
+ * </ul>
+ */
+public class SmithyJavaPlugin implements Plugin<Project> {
```
timocov: Not sure if this should be part of this plugin, but one thing I found annoying and rather distracting is that the output dest folder isn't being cleaned up before running the codegen which might and often does have side-effects (e.g. the code still compiles when it shouldn't, or it doesn't compile when it should) because of left-overs from previous codegen runs. In our plugin extension (based on smithy-base/jar) we have set it up so it might be worth considering adding in here as well.

gradle-plugin/src/main/java/software/amazon/smithy/java/gradle/SmithyJavaPlugin.java
```diff
@@ -0,0 +1,255 @@
...
+
+        smithyBuild.withDependencies(deps -> {
+            if (!ext.getAutoAddDependencies().getOrElse(true)) {
+                return;
+            }
+            String version = SmithyJavaVersion.VERSION;
+            Set<String> resolved = modes.get();
+            addIfAbsent(deps, project.getDependencies(), "codegen-plugin", version);
+            if (resolved.contains("client")) {
+                addIfAbsent(deps, project.getDependencies(), "client-core", version);
```
timocov, line 123: does it need to be here? 🤔

gradle-plugin/src/main/java/software/amazon/smithy/java/gradle/SmithyJavaPlugin.java
```diff
@@ -0,0 +1,255 @@
...
+            }
+            if (resolved.contains("server")) {
+                addIfAbsent(deps, project.getDependencies(), "server-api", version);
```
timocov, line 126: shouldn't this be part of `implementation` instead? 🤔

mtdowling (approved):

gradle-plugin/build.gradle.kts
adwsingh: Thanks, fixed!

gradle-plugin/src/main/java/software/amazon/smithy/java/gradle/SmithyJavaPlugin.java
adwsingh: Nice idea, fixed!

gradle-plugin/src/main/java/software/amazon/smithy/java/gradle/SmithyJavaPlugin.java
adwsingh: `codegen-plugin` only has `client-core` as `compileOnly`, and client mode validates `Client` is present at runtime via Class.forName, so it does need to be present on the smithyBuild classpath. The dependency in the other block is for compiling/consuming the generated source.

gradle-plugin/src/main/java/software/amazon/smithy/java/gradle/SmithyJavaPlugin.java
adwsingh: `Same rationale as the client-core comment`

gradle-plugin/src/main/java/software/amazon/smithy/java/gradle/SmithyJavaPlugin.java
timocov: Yeah this makes sense, but isn't it that for the clients `client-core` must be an `api`-level dependency as the codegen result depends on it and exposes some part of it? `smithyBuild` inherits `implementation` that inherits `api` iirc.

gradle-plugin/src/main/java/software/amazon/smithy/java/gradle/SmithyJavaPlugin.java
rschmitt: `JavaBasePlugin.class` maybe?

gradle-plugin/src/main/java/software/amazon/smithy/java/gradle/SmithyJavaPlugin.java
adwsingh: Yeah, that was my first thought too, but I checked the actual configuration with dependencyInsight and smithyBuild isn’t picking this up from api/implementation in this build. It looks like smithyBuild is on a separate path via compileOnly, so this needs to be present both places.

timocov: Hm, that's weird. This is how I set it up in our plugin and everything looks/works good (we only add `client-core` to `api` for the client, `server-core` to `implementation` for the server and `codegen-plugin` to `smithyBuild` for both) 🤔 We're on v1.4.0 of smithy gradle plugins.

> I checked the actual configuration with dependencyInsight and smithyBuild isn’t picking this up from api/implementation in this build

I'm curious if smithy-base plugin/task actually does some magic internally to make it work. I just tested our plugin it with removing all our specifics and leaving just "`client-core` to `api` and `codegen-plugin` to `smithyBuild`" and it seems works fine - `smithyBuild` succeeds for the client codegen and I can confirm that I don't see `client-core` being added to `smithyBuild` configuration (neither explicitly nor implicitly), but it doesn't complain ¯\_(ツ)_/¯

timocov: @adwsingh if you don't add it to the `smithyBuild`, are you getting errors during the build? If you can share how you reproduce it, I'm happy to check it further from my side.

gradle-plugin/src/main/java/software/amazon/smithy/java/gradle/SmithyJavaPlugin.java
timocov: In some cases (e.g. for the client) it makes sense to add java-library because it will create `api` dependency configuration and for clients that are considered to be published adding smithy-java core dependencies should go to `api` (rather than `implementation`). But for the server, that you usually don't publish anywhere, I believe usually `java` is being used (and it doesn't have `api` configuration). So the plugin should depend on the mode I guess - `java-library` for the client/types, `java` for the server. But even here, in some cases you don't want generated client being published, just having it as a separate module in gradle projects and `java` should work perfectly fine in this case.

So that being said, probably instead of adding `java`/`java-library` it might be better to leave this up to a consumer, and here just react on what they have added:
- if we're building client/types and java plugin was added - then add dependencies to `implementation`
- if we're building client/types and java-library plugin was added - then add dependencies to `api`
- if we're building server and java/java-library plugins were added - then add dependencies to `implementation`
- if we're building for any more but none of the plugins were added - fail the build

In this case the choice is on the consumer, but the plugin makes correct decisions based on the environment it is in imo.

gradle-plugin/src/main/java/software/amazon/smithy/java/gradle/SmithyJavaPlugin.java
adwsingh: My bad, the mistake was using `dependencyInsight` on `smithyBuild` as proof of the whole task classpath. It only proves what is in the `smithyBuild` configuration, not everything the smithyBuild task runs with.

timocov (approved):

gradle-plugin/src/main/java/software/amazon/smithy/java/gradle/SmithyJavaPlugin.java
adwsingh: Turns out this is actually required. What made this confusing is that smithy-base wires more than just the `smithyBuild` configuration into the task. What exposed it is non-main source sets: `jmhSmithyBuild` does not get deps from main `smithyBuild`, so the mode-specific deps need to be added to that target source set’s Smithy config.

timocov: > jmhSmithyBuild does not get deps from main smithyBuild, so the mode-specific deps need to be added to that target source set’s Smithy config.

I'm curios if this will hide potential issues (e.g. the build succeeds and you can publish the result, but consumers won't be able to use it without tweaking because some of the dependencies weren't set correctly). But I guess that's fine for now, it shouldn't cause too much harm anyway.

timocov (approved):

sugmanue (approved):

yasmewad (approved):

PR #1245, opened 2026-06-16
Emit empty payload for RPC protocol event encoding

sugmanue:
### What behavior changes?
RPC protocols (awsJson, rpcv2-cbor) now emit an empty payload (e.g., `{}`) and set the content-type header for events with no payload members. REST protocol (restJson) continue to omit the body and content-type as before.

RPC protocols must always serialize a body even when there are no payload members, unlike REST protocols which omit it.

### How was this validated?
* Existing unit tests pass. Compilation verified across all affected modules.
* Tested against the internal server using REST and RPCv2

- The consolidated condition in `EventSerializer.writeStruct` (`emitEmptyPayload || hasPayloadMembers`)
- Correct `true`/`false` values passed at each protocol call site

### Additional Links

---

### DIFF
--- a/aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/AwsEventEncoderFactory.java
+++ b/aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/AwsEventEncoderFactory.java
@@ -23,6 +23,7 @@ public final class AwsEventEncoderFactory implements EventEncoderFactory<AwsEven
     private final Schema schema;
     private final Codec codec;
     private final String payloadMediaType;
+    private final boolean emitEmptyPayload;
     private final Function<Throwable, EventStreamingException> exceptionHandler;
 
     private AwsEventEncoderFactory(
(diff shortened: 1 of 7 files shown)


aws/aws-event-streams/src/main/java/software/amazon/smithy/java/aws/events/AwsEventEncoderFactory.java
```diff
@@ -45,19 +48,22 @@ private AwsEventEncoderFactory(
...
      * @param payloadMediaType The payload media type
+     * @param emitEmptyPayload Whether to emit an empty payload when there are no payload members
      * @param exceptionHandler The handler to convert exceptions for event streaming
      * @return A new event encoder factory
      */
     public static AwsEventEncoderFactory forInputStream(
             ApiOperation<?, ?> operation,
             Codec codec,
             String payloadMediaType,
+            boolean emitEmptyPayload,
```
mtdowling, line 81: Breaking change unless you add a backwards compatible shim

mtdowling (approved):

PR #1247, opened 2026-06-16
Coerce integer-encoded CBOR values when reading float/double

mtdowling:
readDouble rejected any token that wasn't FLOAT, so a whole-valued double or float encoded as a CBOR integer (e.g. 1.0 written as the major-type-0 integer 1) threw a SerializationException.

### What behavior changes?

### How was this validated?

### Additional Links

---

### DIFF
--- a/codecs/cbor-codec/src/main/java/software/amazon/smithy/java/cbor/CborDeserializer.java
+++ b/codecs/cbor-codec/src/main/java/software/amazon/smithy/java/cbor/CborDeserializer.java
@@ -559,10 +559,27 @@ final class CborDeserializer implements ShapeDeserializer {
     }
 
     private double readDouble(String type, byte token) {
-        if (token != Token.FLOAT) {
+        if (token == Token.FLOAT) {
+            return readDouble(token);
+        } else if (token == Token.POS_INT) {
+            // Tolerate an integer-encoded value for a floating-point member: a whole-valued double (e.g. 1.0)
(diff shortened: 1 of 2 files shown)


codecs/cbor-codec/src/main/java/software/amazon/smithy/java/cbor/CborDeserializer.java
```diff
@@ -559,10 +559,16 @@ public double readDouble(Schema schema) {
...
-            throw badType(type, token);
+        if (token == Token.FLOAT) {
+            return readDouble(token);
+        }
+        // Tolerate an integer-encoded value for a floating-point member: a whole-valued double (e.g.
+        // 1.0) is commonly written as a CBOR integer by other encoders, so coerce instead of rejecting.
+        // token <= NEG_INT is exactly the integer-token set (POS_INT=0, NEG_INT=1), as in readLong.
+        if (token <= Token.NEG_INT) {
+            return readLong(type, token);
         }
```
adwsingh: Its a little more complicated because BigInts can also be encoded as longs, see 

https://github.com/smithy-lang/smithy-java/blob/e19d5a202fc03d211386efeaba64e65aac1b1119/codecs/cbor-codec/src/main/java/software/amazon/smithy/java/cbor/CborReadUtil.java#L106-L107

adwsingh (approved):

PR #1259, opened 2026-06-22
Add S3 plugin and bytecode input capture

mtdowling:
### What behavior changes?

The S3 plugin removes the bucket from the path of an operation if the operation gets the path from the rules engine. The rulesengine also now has the ability to set the input params to a context key when you opt-in to collecting them. Both are needed to power things like introspection and dynamic clients.

```java
// Shared, sink. Samples ~1% of resolutions and logs the decision path.
final class SamplingTraceSink implements BddTraceSink {
    private final EndpointBddTrait trait;

    SamplingTraceSink(EndpointBddTrait trait) {
        this.trait = trait;
    }

    @Override
    public BddTrace begin(Bytecode bytecode, Map<String, Object> params) {
        if (ThreadLocalRandom.current().nextInt(100) != 0) {
            return null; // not sampled: resolution runs the normal, untraced fast path
        }

        // `params` is a live view; copy it if you need it after the callbacks return.
        var record = new StringBuilder("endpoint trace inputs=" + new LinkedHashMap<>(params) + "\n");
        return new BddTrace() {
            @Override
            public void condition(int conditionId, boolean satisfied, boolean branch) {
                record.append("  cond[").append(conditionId).append("] ")
                        .append(trait.getConditions().get(conditionId))   // decode id -> condition fn
                        .append(" => ").append(satisfied)
                        .append(branch == satisfied ? "" : " (complemented)").append('\n');
            }

            @Override
            public void result(int resultId) {
                record.append(resultId < 0
                        ? "  no match\n"
                        : "  result[" + resultId + "] " + trait.getResults().get(resultId) + '\n');
                LOGGER.info(record.toString());
            }
        };
    }
}
```

Attach it per call via the request override (or set it once as a client default):

  ```java
  var override = RequestOverrideConfig.builder()
          .putConfig(RulesEngineSettings.BDD_TRACE_SINK, new SamplingTraceSink(trait))
          .build();
  client.call("GetObject", input, override);
  ```

This is a requirement for calling S3. Getting BDD params is a requirement for introspection.

### How was this validated?

Tests, and calling S3 directly

### Additional Links

---


rulesengine/src/main/java/software/amazon/smithy/java/rulesengine/BytecodeEvaluator.java
```diff
@@ -127,6 +135,58 @@ Endpoint evaluateBdd() {
         return resolveResult(ref - Bdd.RESULT_OFFSET);
     }
 
+    /**
+     * Traced twin of {@link #evaluateBdd()}: same traversal, but reports each step to {@code sink}. Kept
+     * as a separate method so the fast path above stays free of any per-node trace branch; the resolver
+     * picks this only when a sink is present.
+     */
+    Endpoint evaluateBddTraced(BddTraceSink sink) {
```
JordonPhillips, line 143: I'm worried about these two methods de-syncing. Can the existing tests be updated to hit both variants?

JordonPhillips (changes requested):

rulesengine/src/main/java/software/amazon/smithy/java/rulesengine/BytecodeEvaluator.java
mtdowling: Good call. Added tests to ensure they resolve the same.

JordonPhillips (approved):

PR #1260, opened 2026-06-23
Update codegen to support closure-based generation

JordonPhillips:
### What behavior changes?

This updates the code generators to support generating based on shape closures. Shape closures are a new way to define a set of shapes to generate in the model without rooting that set at a service shape.

This has implications that cause some chages throughout the generator. Notably, a service might not exist. There are a number of places that required a service shape to exist that had to be updated to tolerate it being missing. Notably renames may now be driven through the closure definition itself rather than a service shape, so renames needed to be passed through in a lot of generator classes.

Previously types codegen was implemented by creating a synthetic service shape and/or operation and attaching the shapes to it. The generator would then skip those synthetic shapes so that no unwanted scaffolding was generated. That whole system was removed.

Another important implication of shape closures is that a closure can have multiple services. For now, smithy-java is keeping the restriction to have exactly 0 or 1 services. Supporting multiple can be done down the line if there's a need.

Shape codegen is a general feature that Smithy customers have wanted for a long time. Model-defined closures make that possible in a way that shares the definition across generators and removes the need for hacks like the existing synthetic service / operation plumbing.

### How was this validated?

Tests were added, including new integ tests. Existing types codegen codgen settings were kept. 

Existing types codegen behavior should be preserved. There are surely some subtleties of the previous behavior that I've missed, particularly in the mixed mode case.

### Additional Links

This needs https://github.com/smithy-lang/smithy/pull/3156 to build successfully.

---


mtdowling (changes requested):
I think this is just missing an example using the new closure setting

examples/closure-types/smithy-build.json
```diff
@@ -0,0 +1,12 @@
+{
+  "version": "1.0",
+  "plugins": {
+    "java-codegen": {
+      "service": "smithy.example.birds#iBird",
+      "namespace": "software.amazon.smithy.java.example.closure",
+      "headerFile": "license.txt",
+      "modes": ["server", "types"],
+      "closure": "smithy.example.birds#fullService"
```
mtdowling, line 9: Does closure apply to all modes?

JordonPhillips: yes

mtdowling (approved):
Assuming the integration tests pass: ship it

PR #1273, opened 2026-06-25
Fix aws-json type deser

mtdowling:
### What behavior changes?

Strip namespaces when deserializing AWS JSON types to make exception handling work

Wasn't working

### How was this validated?

Manually testing against services

### Additional Links

---


aws/client/aws-client-awsjson/src/main/java/software/amazon/smithy/java/aws/client/awsjson/AwsJson1Protocol.java
```diff
@@ -31,8 +31,12 @@ public AwsJson1Protocol(ShapeId service) {
     }
 
     private static ShapeId extractErrorType(Document document, String namespace) {
+        // Per the awsJson1_0 SEP, the __type discriminator may carry an over-the-wire namespace
```
kstich:
```suggestion
        // Per the awsJson1_0 spec, the __type discriminator may carry an over-the-wire namespace
```

kstich (approved):

PR #1275, opened 2026-06-26
Fix blob serialization in MCP output adaptation

DivyaMaddipudi:
### What behavior changes?

`adaptOutputDocument` no longer throws `SerializationException: Expected a blob document, but found string` when a downstream RPC v2 response contains a Blob field that arrives as an already-base64-encoded string Document. Instead, it passes the string through unchanged. Binary blob Documents continue to be base64-encoded as before.

When MCP proxies a `tools/call` to a downstream Smithy service using RPC v2 CBOR, Blob fields in the response can be deserialized as string Documents (already base64-encoded) rather than typed blob Documents. The previous code unconditionally called `doc.asBlob()`, which throws when the document type is STRING.

### How was this validated?
- Unit tests added in `McpServiceTest` covering all three branches: STRING passthrough, BLOB encode, and unsupported type error
- Structure-level test validating nested blob fields within a structure are adapted correctly
- Full `McpServerTest` suite passes
- End-to-end validation against a live MCP server

- `McpService.java` lines 1319–1324: the new nested switch on `doc.type()` inside the `BLOB` case
- `McpServiceTest.java`: verify the test coverage is sufficient and the package-private visibility is acceptable

### Additional Links

  ---

### DIFF
--- a/mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpService.java
+++ b/mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpService.java
@@ -1308,7 +1308,7 @@ public final class McpService {
         return Document.of(EPOCH_SECONDS.readFromNumber(doc.asNumber()));
     }
 
-    private Document adaptOutputDocument(Document doc, Schema schema) {
+    Document adaptOutputDocument(Document doc, Schema schema) {
         if (doc == null) {
             return null;
         }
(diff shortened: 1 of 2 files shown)


mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/McpService.java
```diff
@@ -1308,15 +1308,20 @@ private static Document adaptTimestamp(Document doc) {
...
         return switch (toType) {
             case BIG_DECIMAL -> Document.of(doc.asBigDecimal().toString());
             case BIG_INTEGER -> Document.of(doc.asBigInteger().toString());
-            case BLOB -> Document.of(Base64.getEncoder().encodeToString(ByteBufferUtils.getBytes(doc.asBlob())));
+            case BLOB -> switch (doc.type()) {
+                case STRING -> Document.of(doc.asString());
+                case BLOB -> Document.of(
+                        Base64.getEncoder().encodeToString(ByteBufferUtils.getBytes(doc.asBlob())));
+                default -> badType(doc.type(), toType);
+            };
```
adwsingh, line 1324: This is a incorrect fix and in the wrong layer. The premise of this fix is also incorrect

> to a downstream Smithy service using RPC v2 CBOR, Blob fields in the response can be deserialized as string Documents (already base64-encoded) rather than typed blob Documents.

That's the representation over the wire not after materialization.

DivyaMaddipudi: The actual issue is Document.of(output) at McpService.java:704 round-trips JsonDocuments.StringDocument (has asBlob()) into core Documents.StringDocument (throws on asBlob()) before adaptOutputDocument resolves @oneOf document members containing blob fields. Should I fix it by skipping the round-trip  (cast StructDocument directly since it implements Document), or add asBlob() to core Documents.StringDocument?

PR #1289, opened 2026-07-10
Fixed deserialization of string httpPayload when the response is readonly buffer

timocov:
### What behavior changes?

Fixed a bug in deserialising string from a readonly buffer.

Explain the motivation: bug, feature request, refactor, performance, etc

When an output has a string as a httpPayload

```smithy
@output
structure OperationOutput {
    @httpPayload
    body: String
}
```

deserialization of such response fails with:

```
    software.amazon.smithy.java.client.core.error.TransportException
        Caused by:
        java.nio.ReadOnlyBufferException
            at java.base/java.nio.ByteBuffer.arrayOffset(ByteBuffer.java:1535)
            at software.amazon.smithy.java.http.binding.PayloadDeserializer.readString(PayloadDeserializer.java:123)
```

After debugging it turned out that the buffer passed to the deserializer is readonly thus accessing `arrayOffset` is not possible and results in the error. Applying similar logic as in https://github.com/smithy-lang/smithy-java/blob/4ac99cca61c80cfa21e79fbf65b44470d6514548/io/src/main/java/software/amazon/smithy/java/io/datastream/ByteBufferDataStream.java#L64-L70

### How was this validated?

Added tests, tested on local models.

### Additional Links

---

### DIFF
--- a/http/http-binding/src/main/java/software/amazon/smithy/java/http/binding/PayloadDeserializer.java
+++ b/http/http-binding/src/main/java/software/amazon/smithy/java/http/binding/PayloadDeserializer.java
@@ -120,9 +120,13 @@ final class PayloadDeserializer implements ShapeDeserializer {
         }
 
         var buffer = body.asByteBuffer();
-        int pos = buffer.arrayOffset() + buffer.position();
-        int len = buffer.remaining();
-        return new String(buffer.array(), pos, len, StandardCharsets.UTF_8);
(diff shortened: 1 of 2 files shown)


mtdowling (approved):
Thanks!

timocov:
@mtdowling oops sorry forgot to apply spotless to tests, fixed now 🙏

PR #1290, opened 2026-07-10
Fixed flaky tests when running with UK locale

timocov:
### What behavior changes?

No tests failures when running with non-US locale.

Apparently September has different short versions depending on the locate you're in 😄 To reduce tests flakiness.

### How was this validated?

Run tests.

### Additional Links

https://stackoverflow.com/questions/69267710/septembers-short-form-sep-no-longer-parses-in-java-17-in-en-gb-locale

---


mtdowling (approved):
whoops! Thanks!

PR #1292, opened 2026-07-13
Fixed deserialising string/blob payloads when transferred with chunked-encoding

### LINKED ISSUE #1291: Chunked-encoded string/blob payload is being ignored
When declaring an operation with a string (or blob I guess as well) http payload:

```smithy
@http(method: "GET", uri: "/string")
operation GetString {
    output := {
        @httpPayload
        data: String
    }
}
```

And the server responds with [chunked encoding](https://en.wikipedia.org/wiki/Chunked_transfer_encoding) (e.g. because it doesn't know the size of the response ahead of time), the generated client will silently ignore the response.

This happens because the content length isn't know ahead of time and [`body.contentLength()`](https://github.com/smithy-lang/smithy-java/blob/d73130e343fa00bfa98629949df44a3c50b102ac/http/http-binding/src/main/java/software/amazon/smithy/java/http/binding/HttpBindingDeserializer.java#L238) returns `-1` (which feels like a valid case?).

Not sure if that check is really needed there? I tried to remove it locally and it parses the response correctly, but not too sure if there are other implications.

timocov:
Fixes #1291

### What behavior changes?

Fixed handling string/blob http payloads with chunked encoding.

Negative (-1) content length is a valid value which indicates that the value just isn't known ahead of time. If a server responds with chunked encoding, the value will not be known ahead of time and it will result in the response being ignored (only when http payloads on string/blob was being used).

### How was this validated?

Run tests, built and test locally.

I guess it might be possible that whilst the content length is not known ahead of time but it resolves to empty response in the end it might cause some issues?

### Additional Links

---

### DIFF
--- a/http/http-binding/src/main/java/software/amazon/smithy/java/http/binding/HttpBindingDeserializer.java
+++ b/http/http-binding/src/main/java/software/amazon/smithy/java/http/binding/HttpBindingDeserializer.java
@@ -235,7 +235,7 @@ final class HttpBindingDeserializer extends SpecificShapeDeserializer implements
             if (bb.remaining() > 0) {
                 structMemberConsumer.accept(state, member, payloadCodec.createDeserializer(bb));
             }
-        } else if (body != null && body.contentLength() > 0) {
+        } else if (body != null && body.contentLength() != 0) {
             structMemberConsumer.accept(state, member, new PayloadDeserializer(payloadCodec, body));
         }
     }


timocov:
Alternatively, how about something like https://github.com/smithy-lang/smithy-java/commit/c9c790183ff975c56e981eec0d81dff74721f3db instead?

mtdowling (approved):

mtdowling:
Thanks! Good catch

PR #1304, opened 2026-08-04
Add MCP protocol version 2025-11-25

DivyaMaddipudi:
### What behavior changes?

**Before:** When a client sends `"protocolVersion": "2025-11-25"` in the initialize request, the server treats it as `UnknownVersion` and omits `protocolVersion` from the response (falling back to the builder default `"2024-11-05"`). This breaks the handshake for clients expecting the server to echo back `"2025-11-25"`.

**After:** The server recognizes `"2025-11-25"` as a known version and correctly echoes it back in the initialize response.

Clients like Claude Code and Cursor have adopted MCP spec version `2025-11-25` and are sending it during initialization. The smithy-java MCP server only recognized versions up to `2025-06-18`, causing initialization failures for these clients. The `2025-11-25` spec uses the same handshake model (initialize → response → initialized) as prior versions, so no protocol-level changes are required beyond accepting the version string.

### How was this validated?

- New `ProtocolVersionTest`: validates version resolution, null handling, ordering, and unknown-version behavior
- New `McpServerTest.initializeWithV2025_11_25ProtocolVersion`: end-to-end test confirming the server echoes the version and serves tools/list after handshake
- All 87 existing tests continue to pass

- `ProtocolVersion.java`: The only production change — adds `v2025_11_25` to the sealed class hierarchy and switch statement. Note that `defaultVersion()` intentionally remains at `v2025_03_26` until the new capabilities (tasks, richer elicitation) are implemented.

### Additional Links

- [MCP spec 2025-11-25](https://modelcontextprotocol.io/specification/2025-11-25)
- [MCP versioning docs](https://modelcontextprotocol.io/docs/2026-07-28/learn/versioning)

### DIFF
--- a/mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/ProtocolVersion.java
+++ b/mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/ProtocolVersion.java
@@ -10,7 +10,15 @@ import software.amazon.smithy.utils.SmithyUnstableApi;
 @SmithyUnstableApi
 public abstract sealed class ProtocolVersion implements Comparable<ProtocolVersion>
         permits ProtocolVersion.UnknownVersion, ProtocolVersion.v2024_11_05, ProtocolVersion.v2025_03_26,
-        ProtocolVersion.v2025_06_18 {
+        ProtocolVersion.v2025_06_18, ProtocolVersion.v2025_11_25 {
+    public static final class v2025_11_25 extends ProtocolVersion {
+        public static final v2025_11_25 INSTANCE = new v2025_11_25();
+
+        private v2025_11_25() {
+            super("2025-11-25");
+        }
+    }
+
     public static final class v2025_06_18 extends ProtocolVersion {
         public static final v2025_06_18 INSTANCE = new v2025_06_18();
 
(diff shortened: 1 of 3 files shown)


mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/ProtocolVersion.java
```diff
@@ -10,7 +10,15 @@
 @SmithyUnstableApi
 public abstract sealed class ProtocolVersion implements Comparable<ProtocolVersion>
         permits ProtocolVersion.UnknownVersion, ProtocolVersion.v2024_11_05, ProtocolVersion.v2025_03_26,
-        ProtocolVersion.v2025_06_18 {
+        ProtocolVersion.v2025_06_18, ProtocolVersion.v2025_11_25 {
+    public static final class v2025_11_25 extends ProtocolVersion {
```
franco2002lu, line 14: do you also intend to implement some of the new mcp features as well?

franco2002lu (approved):

mcp/mcp-server/src/main/java/software/amazon/smithy/java/mcp/server/ProtocolVersion.java
DivyaMaddipudi: For now we would like to support only the protocol version

PR #1309, opened 2026-08-06
Add config settings to customize the input and output suffixes

sugmanue:
### What behavior changes?

This change allows the users to customize the input and output suffixes using the plugin settings.

### How was this validated?

### Additional Links

---


joewyz (approved):

codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/JavaCodegenSettings.java
```diff
@@ -311,6 +320,14 @@ public ObjectNode httpConfig() {
         return httpConfig;
     }
 
+    public String inputSuffix() {
+        return inputSuffix;
+    }
+
+    public String outputSuffix() {
+        return outputSuffix;
+    }
```
timocov, line 329: It would be nice to have an option to set them via codegen integration as well (similar to default plugins/settings), do you think this is something the team can expose as an option? 🤔

sugmanue: Sure, I'll take a look at it after this upcoming release.

PR #1311, opened 2026-08-07
Update closure types example

JordonPhillips:
### What behavior changes?

This updates the closure types example to show not only combined mode, but also types only and no types. Having every type in one example is useful to show how the same model may be used in multiple ways. This example will be referenced in a blog post announcing closure-based codegen, so keeping it all in one is particularly important.

It showcases the feature more broadly. And conforms to what I want to discuss in a blog post discussing the feature.

### How was this validated?

There's tests. Even tests covering example code, to an extent.

Focus on the model and the readme.

### Additional Links

---


examples/closure-types/README.md
```diff
@@ -20,3 +63,12 @@ To use this example as a template, run the following command with the
 ```console
 smithy init -t closure-types --url git@github.com:smithy-lang/smithy-java.git
 ```
+
+Then build it and run the tests:
```
joewyz, line 67: Tried to init this locally and it failed due to smithyjava version setting [here](https://github.com/smithy-lang/smithy-java/blob/d7ce5ebca272dfa09126e46e6b84257e135ca0f2/examples/gradle.properties#L1) can you please also update it to match our current version?

JordonPhillips: O dang you right. It's been that way forever. The gradle plugin is old as heck too

joewyz (approved):

PR #1314, opened 2026-08-07
Fixed bug when validator of unions produces incorrect path

timocov:
### What behavior changes?

Fixed a bug, when a validator is applied and one of path segments represents a union targetting a structure, instead of providing its member name will produce something like:

```
/Schema{id='smithy.example#U$d', type=structure}/foo`
```

instead of

```
/d/foo
```

Fixing a bug.

### How was this validated?

Added tests.

Trivial change, no special requests here.

### Additional Links

N/A

---


core/src/main/java/software/amazon/smithy/java/core/schema/ValidatorOfUnion.java
```diff
@@ -197,7 +197,7 @@ public <T> void writeMap(Schema member, T state, int size, BiConsumer<T, MapSeri
 
     @Override
     public void writeStruct(Schema member, SerializableStruct struct) {
-        validator.pushPath(member);
+        validator.pushPath(member.memberName());
```
timocov, line 200: This is the main fix, and it feels like it was possible because the API was too loose (it accepted `Object`) so I made it more strict and now it only accepts strings (and for swapping the path it accepts numbers), but the underlying container is still `Object[]` so that we can accept both `String` and `int` (and `null`). That API is not public nor it is exposed, but if prefer to revert method signature change let me know, happy to revert it.

core/src/test/java/software/amazon/smithy/java/core/schema/ValidatorTest.java
```diff
@@ -537,6 +541,24 @@ public void ignoresNullValuesInUnion() {
...
+    @Test
+    public void producesCorrectPathInSubObjects() {
+        Validator validator = Validator.builder().build();
+        var unionSchema = getTestUnionSchema();
+
+        var errors = validator.validate(s -> {
+            s.writeStruct(unionSchema, TestHelper.create(unionSchema, (schema, serializer) -> {
+                serializer.writeStruct(schema.member("d"),
+                        TestHelper.create(schema.member("d"), (passedSchema, innerSerializer) -> {
+                            serializer.writeString(passedSchema.member("foo"), "1234");
```
JordonPhillips: You appear to be using the wrong serializer here

```suggestion
                            innerSerializer.writeString(passedSchema.member("foo"), "1234");
```

JordonPhillips (changes requested):
Looks good, just a small issue with the test

core/src/test/java/software/amazon/smithy/java/core/schema/ValidatorTest.java
timocov: Oops, good catch! Fixed now. Both serializers just happened to do the same when writing a string for a string so it worked haha 😄

JordonPhillips (approved):

PR #1336, opened 2026-08-28
Paginate MCP tools/list and prompts/list

sugmanue:
### What behavior changes?

`McpServerProxy.listTools()` / `listPrompts()` now follow MCP cursor pagination. Before, they issued a single JSON-RPC call and ignored `nextCursor`, so a server that paged its results (commonly ~30 items per page) was silently truncated to its first page everywhere `McpServerProxy` is used. After, they return the union of all pages. Single-page servers are unchanged and still make exactly one round-trip.

Two related fixes ship alongside it:

- The `tools/list_changed` refresh in `McpService` no longer runs inline on the   transport reader thread. Previously that path could deadlock: the refresh   calls `listTools()` whose response is read by the same reader thread. It now   runs on a dedicated single-thread executor.
- `McpService`'s tool/prompt/proxy registries are now thread-safe. They were   mutated by compound (multi-step) operations from several threads with no mutual   exclusion.

Two intentional, minor observable differences reviewers should be aware of:

1. `tools/list` and `prompts/list` now return a stable, deterministic order    (insertion order) instead of an unspecified hash-map order. 
2. `McpService.getProxies()` now returns an immutable snapshot of the proxies at    the time of the call. It previously returned the live internal map; mutating    the returned map now throws, and the snapshot does not reflect proxies added    afterward.

Bug. MCP `tools/list` and `prompts/list` are cursor-paginated, but the client only ever read the first page, so any server or gateway federating more than one page of tools had its tool set truncated with no error and no log, which made it hard to notice.

This also re-lands the pagination fix from #1308 (reverted in #1312) together with a proper fix for the thread-safety concern that caused that revert: moving the `tools/list_changed` refresh off the reader thread requires the registries it mutates to be safe for concurrent access, which they were not.

### How was this validated?

`./gradlew :mcp:mcp-server:build` (unit + integration tests, Spotless, SpotBugs)
green: 200 tests, 0 failures.

Tests added or updated:

- `McpServerProxyTest` (new, 12 cases): multi-page union with exact   `nextCursor` -> `cursor` threading, single-page regression, blank-cursor   termination, repeated-cursor abort, cycling (`A -> B -> A`) abort, page-cap   boundary, error on the first page, error on a later page, null-`result`   handling, `listPrompts` guard path, and immutable return.
- `HttpMcpProxyTest` (+1): a real localhost `HttpServer` pages `tools/list` with   a `nextCursor` and asserts `listTools()` follows it, reading the cursor off the   actually-deserialized response.
- `StdioProxyTest` (new, 1): a silent server times out via the per-request   timeout instead of hanging.
- `McpServiceTest` (new, 3): the `tools/list_changed` refresh runs off the   notifying thread (deadlock regression), add-then-prune adds new tools while   pruning stale ones, and a concurrency stress test (4 adders + 4 refreshers +   4 readers x 200 iterations) asserts no lost updates and no exceptions.
- `McpServerTest`: the cache-invalidation test was updated for the now-async   refresh (it polls for the refresh instead of assuming inline completion). 
- `McpServerIntegrationTest`: helpers that assumed the first listed tool was   `McpEcho` now look it up by name, since tool order is deterministic rather   than hash-based.

- `McpServerProxy#listPaginated(...)` (`mcp/mcp-server/.../McpServerProxy.java`): the pagination loop and its three termination guards (blank/absent cursor ends, repeated or cycling cursor aborts, `maxListPages()` cap of 1000), the null-`result` guard, and the immutable return.
- `mcp/mcp-schemas/model/main.smithy`: the additive optional `nextCursor` on `ListToolsResult` / `ListPromptsResult`.
- `McpService` (`mcp/mcp-server/.../McpService.java`): the copy-on-write registries (`tools` / `prompts` / `proxies` / `services` held as volatile immutable snapshots mutated under a single `registryLock`, with network I/O kept outside the lock), the off-reader-thread `tools/list_changed` refresh with add-then-prune, and `getProxies()` returning a snapshot.
- `StdioProxy#rpc` (`mcp/mcp-server/.../StdioProxy.java`): the builder-configurable per-request timeout (default 5m, symmetric with `HttpMcpProxy`) and pending-request cleanup.

### Additional Links

- MCP specification, Server Utilities: Pagination (cursor-based `nextCursor` / `cursor`): https://modelcontextprotocol.io/specification
- Prior art: #1308 (merged then reverted via #1312 for the thread-safety concern this PR addresses).

---

### DIFF
--- a/mcp/mcp-schemas/model/main.smithy
+++ b/mcp/mcp-schemas/model/main.smithy
@@ -85,6 +85,10 @@ structure ServerInfo {
 
 structure ListToolsResult {
     tools: ToolInfoList
+
+    /// Opaque cursor for the next page of results, per MCP pagination (spec section 5.4).
+    /// Absent when there are no further pages.
+    nextCursor: String
 }
 
 structure ToolInfo {
@@ -219,6 +223,10 @@ structure TextContent {
 
 structure ListPromptsResult {
     prompts: PromptInfoList
+
+    /// Opaque cursor for the next page of results, per MCP pagination (spec section 5.4).
+    /// Absent when there are no further pages.
+    nextCursor: String
 }
 
 list PromptInfoList {
--- a/mcp/mcp-server/src/it/java/software/amazon/smithy/java/mcp/server/McpServerIntegrationTest.java
+++ b/mcp/mcp-server/src/it/java/software/amazon/smithy/java/mcp/server/McpServerIntegrationTest.java
@@ -133,13 +133,18 @@ class McpServerIntegrationTest {
     private ToolSchemas getMcpEchoToolSchemas() {
         write("tools/list", Document.of(Map.of()));
         var responseJson = readRawResponse();
-        var toolNode = OBJECT_MAPPER.readTree(responseJson).path("result").path("tools").get(0);
-        var inputSchemaNode = toolNode.path("inputSchema");
-        var outputSchemaNode = toolNode.path("outputSchema");
-        return new ToolSchemas(
-                SCHEMA_FACTORY.getSchema(inputSchemaNode),
-                SCHEMA_FACTORY.getSchema(outputSchemaNode),
-                toolNode);
+        var toolsNode = OBJECT_MAPPER.readTree(responseJson).path("result").path("tools");
+        for (var toolNode : toolsNode) {
+            if (toolNode.path("name").asString().equals("McpEcho")) {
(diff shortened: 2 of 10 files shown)


adwsingh:
Aggregating tools on the proxy side is not the right design but given we are rewriting all of these parts in #1341 , I am fine to ship this in its current form.

adwsingh (approved):

Issue #1337, opened 2026-08-28
MCP server: support protocol revision 2026-07-28 (stateless / modern era)


psp65:
## Summary

MCP protocol revision [2026-07-28](https://modelcontextprotocol.io/specification/2026-07-28/changelog) makes the protocol stateless: the `initialize` handshake is replaced by per-request versioning in `params._meta`, servers **MUST** implement `server/discover`, version mismatches return `UnsupportedProtocolVersionError` (-32022), every result carries a required `resultType`, and list results carry required `ttlMs`/`cacheScope` caching hints (SEP-2549).

smithy-java's `mcp-server` currently implements the handshake era only (2024-11-05 … 2025-11-25). Python SDK ≥ 2.0.0 clients (released 2026-07-28; the 1.x line is in maintenance mode since then) connect today **only** via their backward-compatibility fallback path — and that fallback currently triggers off an invalid response rather than clean negotiation. Clients pinned to the modern revision, or any strict client, fail outright.

We run a large internal MCP gateway built on this engine (it also powers our Coral-service MCP servers), have validated the gaps end-to-end against Python SDK 2.0.0, and have working, tested patches we'd like to contribute.

## What breaks today

Validated against `mcp==2.0.0` (`mode=auto` / `mode=legacy` / pinned `mode="2026-07-28"`):

1. **Unknown JSON-RPC methods return no response at all** — both dispatch switches in `McpService` end in `default -> null`, so a request with an `id` (including the SDK's `server/discover` probe) gets an empty reply. This violates JSON-RPC, and the V2 SDK's auto-mode fallback only survives it by treating the resulting client-side parse error as legacy evidence. The description of #1304 documents this exact symptom for `initialize`.
2. **`initialize` with an unknown `protocolVersion` answers the *oldest* version** — `handleInitialize` leaves the field unset for unknown versions, so the Smithy model default (`2024-11-05`) leaks into the result. The spec requires answering with the latest supported version. (#1304 fixed one instance by adding 2025-11-25 to the known list; the general fallback remained.)
3. **No `server/discover`, no `-32022`** — modern-era requests carrying `_meta["io.modelcontextprotocol/protocolVersion"]` are served handshake-shaped results, which then fail the SDK's strict wire validation (`ListToolsResult: cacheScope/resultType/ttlMs Field required`) instead of getting a clean version error.
[...]

PR #1339, opened 2026-08-29
Respond with the latest supported protocol version when initialize requests an unknown one

psp65:
> **Note:** stacked on #1338 — the first commit here is that PR's; this PR's own change is the second commit. Kept as draft until #1338 lands.

### What behavior changes?

`initialize` requesting a protocol version the server does not know (e.g. a future revision) now answers with the **latest version the server supports** (`2025-11-25` today):

```json
{"result":{"protocolVersion":"2025-11-25", ...}}
```

Previously `handleInitialize` left the field unset for unknown versions, so the Smithy model default — `2024-11-05`, the *oldest* version — leaked into `InitializeResult`. Known requested versions are still echoed unchanged, and the behavior for a missing `protocolVersion` param is intentionally untouched (pinned by an existing test).

The MCP spec requires that a server that does not support the requested version respond with the latest version it does support. #1304's description documents this exact bug; it was fixed for `2025-11-25` by adding that version to the known list, which left the general unknown-version fallback broken. This is the general fix, and step 2 of the 2026-07-28 ladder in #1337.

### How was this validated?

- `ProtocolVersionTest.latestVersionIs2025_11_25` and `McpServerTest.initializeWithUnknownProtocolVersionAnswersLatestSupported` (initialize @`2026-07-28` → `"2025-11-25"`); known-version echo was already covered by existing tests. Module `test`/`check` green, including the integration suite.
- End-to-end against our internal gateway: Python SDK 1.28.1 and 2.0.0 (`auto` and `legacy` modes) negotiate `2025-11-25` exactly as before.

- `ProtocolVersion.latestVersion()` is derived from the supported-version registry so it cannot drift when future versions are added — and the registry moved into an initialization-on-demand holder. That holder is load-bearing: computing the max during class init of the sealed base class NPEs when a subclass `INSTANCE` is the first member of the hierarchy touched (base `<clinit>` runs while the subclass clinit is in progress and reads a still-null `INSTANCE`). The integration suite caught this; the unit suite alone would not have.
- The `UnknownVersion` branch in `handleInitialize`.

### Additional Links

- #1337 (MCP 2026-07-28 support proposal; step 2)
- #1304 (documents the symptom; fixed one instance)
- Stacked on #1338

---


adwsingh (approved):

psp65:
Closing as superseded: #1341 (merged as 6e5c8baa3) implements this behavior — initialize with an unknown or modern requested version now answers the latest supported handshake version via the protocol registry's initialization fallback, which we verified on the merged code (initialize @2026-07-28 → 2025-11-25). The file this PR patches was deleted by the refactor, so it can no longer merge. Thanks for the earlier review!

PR #1344, opened 2026-08-30
Serialize structure members in schema order, not declaration order

adwsingh:
### What behavior changes?
Serialize structure members in schema order, not declaration order

So that adding changing the order of fields in the model doesn't change the order of serialization.

### How was this validated?

### Additional Links

---

### DIFF
--- a/codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/generators/StructureSerializerGenerator.java
+++ b/codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/generators/StructureSerializerGenerator.java
@@ -55,7 +55,7 @@ record StructureSerializerGenerator(
     private void writeMemberSerialization(JavaWriter writer) {
         boolean isError = shape.hasTrait(ErrorTrait.class);
 
-        for (var member : shape.members()) {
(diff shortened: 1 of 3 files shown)


codegen/codegen-core/src/main/java/software/amazon/smithy/java/codegen/generators/StructureSerializerGenerator.java
```diff
@@ -55,7 +55,7 @@ public void serializeMembers(${shapeSerializer:T} serializer) {
     private void writeMemberSerialization(JavaWriter writer) {
         boolean isError = shape.hasTrait(ErrorTrait.class);
 
-        for (var member : shape.members()) {
+        for (var member : CodegenUtils.getSortedMembers(shape)) {
```
mtdowling, line 58: Does this alphanumerically sort structure and union members? Or does it return those in model order / idx order if available?

adwsingh: No, it just ensures that non-nullable fields come before nullable ones.

mtdowling (approved):

